PR #9, opened 2025-10-08
feat: kernels check CI.

sayakpaul:
Is this what you were envisioning?

### DIFF
--- a/scripts/run_kernels_checks.py
+++ b/scripts/run_kernels_checks.py
@@ -1,0 +1,104 @@
+import argparse
+import logging
+import os
+import subprocess
+
+ORG = "kernels-community"
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser(
+        description=("Run `kernels check` for every top-level directory in the current repository.")
+    )
+    parser.add_argument(
+        "--root",
+        default=".",
+        help="Local path whose immediate subdirectories will be checked. Defaults to the current directory.",
+    )
+    parser.add_argument(
+        "--exclude",
+        action="append",
+        default=[".github", "scripts"],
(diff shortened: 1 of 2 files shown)


sayakpaul:
Hmm, the yml workflow file wasn't automatically added when I did `git add .`. Checking.

danieldk:
Can we do a test run from this PR? (not sure)

.github/workflows/nightly_kernel_check.yml
```diff
@@ -0,0 +1,28 @@
...
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install git+https://github.com/huggingface/kernels kernel-abi-check
```
sayakpaul, line 24: `kernels check` requires the `kernels` installation to be from `main`. Okay for now I guess?

danieldk: Yeah, I don't think that's a problem.

scripts/run_kernels_checks.py
```diff
@@ -0,0 +1,104 @@
...
+    filtered = {exclude.strip() for exclude in excludes}
+    try:
+        entries = os.listdir(root)
+    except OSError as err:
+        raise RuntimeError(f"Unable to list directories under {root}: {err}") from err
+
+    directories = []
+    for entry in sorted(entries):
+        if entry.startswith(".") or entry in filtered:
+            continue
```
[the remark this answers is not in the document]
sayakpaul: Oh then I think we're good to go!

PR #31, opened 2025-10-13
mra: add kernel for transformers

MekkCyber:


drbh (approved):
lgtm

*built and test locally and added a readme script + CI is passing once the flake was bumped

PR #45, opened 2025-10-15
non-intrusive changes for ci.

sayakpaul:


danieldk:
> Is this expected given we have already released that feature in the stable?

Yes, kernel-builder still has the previous version, currently updating.

danieldk:
Should work on `main` now.

PR #63, opened 2025-11-03
Changing output to follow Transformers v5

shawntan:
Modifying kernel to match transformers v5 based on: https://github.com/huggingface/transformers/issues/40822#issuecomment-3437837982

### DIFF
--- a/scattermoe/torch-ext/scattermoe/layers.py
+++ b/scattermoe/torch-ext/scattermoe/layers.py
@@ -48,5 +48,5 @@ class ScatterMoEGatedMLP(nn.Module):
             gates=routing_weights
         )
         layer_output = layer_output.view(bsz, length, emb_size)
-        return layer_output, router_logits
+        return layer_output
 


scattermoe/torch-ext/scattermoe/layers.py
```diff
@@ -48,5 +48,5 @@ def forward(self, layer_input):
             gates=routing_weights
         )
         layer_output = layer_output.view(bsz, length, emb_size)
-        return layer_output, router_logits
+        return layer_output
```
MekkCyber: so we don't need the router_logits in the kernel ?

MekkCyber (commented):
Thanks!

scattermoe/torch-ext/scattermoe/layers.py
shawntan: Ok, Mamba's been updated, I've tested it. Sorry about that!

PR #70, opened 2025-11-17
flash-attn2: [XPU] Add Local Mask Feature

YangKai0616:
This PR adds the `local mask` feature to XPU and enables the related UTs. 
Additionally, it fixes the library name (`flash_attn -> flash_attn2`) in test file.

I have successfully built this PR locally using nix.


YangKai0616:
@drbh @danieldk , please help review. Thanks!

danieldk (approved):

danieldk:
The build should get uploaded automatically, but if it times out again, I'll do a manual upload.

PR #82, opened 2025-12-05
make ScatterMoe more easily patchable for HF MoE modeling

winglian:
This PR makes it easier to simply use something like the snippet below for a drop in replacement for all the MoEs based on qwen2 in transformers v5. The original `ScatterMoEGatedMLP` isn't quite usable since it relies on `router` instead of `gate` and `input_linear`, `output_linear` instead of `gate_up_proj`, `down_proj` in the v5 modeling. 

```python
register_kernel_mapping({
    "HFScatterMoEParallelExperts": {
        "cuda": {
            Mode.TRAINING: LayerRepository(
                repo_id="axolotl-ai-co/scattermoe",
                layer_name="HFScatterMoEGatedMLP",
            ),
            Mode.INFERENCE: LayerRepository(
                repo_id="axolotl-ai-co/scattermoe",
                layer_name="HFScatterMoEGatedMLP",
            ),
        },
    }
})
replace_kernel_forward_from_hub(Qwen2MoeSparseMoeBlock, "HFScatterMoEParallelExperts")
```

### DIFF
--- a/scattermoe/torch-ext/scattermoe/layers.py
+++ b/scattermoe/torch-ext/scattermoe/layers.py
@@ -50,3 +50,102 @@ class ScatterMoEGatedMLP(nn.Module):
         layer_output = layer_output.view(bsz, length, emb_size)
         return layer_output
 
+
+
+
+class HFScatterMoEGatedMLP(nn.Module):
+    """
+    ScatterMoE-accelerated forward pass for HF MoEs based on Qwen2MoE.
+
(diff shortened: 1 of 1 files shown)


sayakpaul:
Is there still an issue?

vasqu:
Also following this thread now :hugs: is there anything blocking use @winglian @shawntan? If it's only for training we could also restrict that for now, I think it could be very valuable for us on transformers

PR #86, opened 2025-12-09
add fp8-fbgemm kernels

sywangyi:
@danieldk @MekkCyber please help review


MekkCyber (commented):
Thanks for the quick action @sywangyi

PR #92, opened 2025-12-15
flash-attn2: [XPU] Refactor FlashAttention with new Xe atoms

YangKai0616:
I need to refine the code before merging.


YangKai0616:
@drbh @danieldk , we can start the review now. Thanks.

PR #95, opened 2025-12-16
Add PR_TITLE env variable to validation step

paulinebm:
Set PR_TITLE environment variable for kernel validation.


danieldk (approved):
Thanks!

PR #242, opened 2026-01-19
rotary: Remove The Unused Parameter

YangKai0616:
Remove the unused `position_ids` parameter to adapt to the latest `transformers`.


YangKai0616:
Hey @danieldk , just like [here](https://github.com/huggingface/transformers/blob/314f10929a2215b74c2ad6ecf7b2f380c9b7468a/src/transformers/models/llama/modeling_llama.py#L145), we need to remove the extra parameters to match the kernels' replacement. Please help review, thank you!

PR #270, opened 2026-01-22
rmsnorm: fix crash if work with torch.compile

sywangyi:


sywangyi:
@danieldk please help review. fix crash if work with torch.compile and add test case.

PR #358, opened 2026-02-11
flash-attn2: [XPU] Rename cutlass_sycl to sycl_tla

YangKai0616:
Reference PR https://github.com/huggingface/kernels/pull/277. Rename `cutlass-sycl` to `sycl-tla`.
We need to wait for the `kernels` to be updated before activating this PR.


YangKai0616:
> Merged in `kernels`, so should work after a `nix flake update`.

Done! It can build and test successfully on my local machine.

PR #359, opened 2026-02-11
flash-attn2: Fix cpu flash-attn2 fallback path

jiqing-feng:
Fallback path cannot use brgemm. It will cause core dump if no amx detected.


danieldk (approved):
Looks good, merged in #360.

PR #365, opened 2026-02-12
Add Apache 2.0 LICENSE

jp-gr:
https://github.com/huggingface/kernels already uses Apache 2.0 so I assume this is an appropriate choice?


MekkCyber (commented):
Thank you @jp-gr for the pr

danieldk:
Thanks again for the PR! Closing this, since we now have a license attribute on each kernel. This will also be rendered on the Hub.

PR #530, opened 2026-04-07
finegrained-fp8: fused moe kernels

IlyasMoutawwakil:
moe_tflops_grouped
moe_tflops_batched
faster and more accurate fused fp8 for both long context (grouped) and decode (batched)


sayakpaul:
@IlyasMoutawwakil anything blocking merge here?

IlyasMoutawwakil:
@sayakpaul left unfinished while working on sonicmoe and deepgemm v1/v2, coming back to it after #954 
will try to find some sort simple way to enable the fusion on all the quant recipes we support (we have too many now).

PR #531, opened 2026-04-07
sonic-moe: Add sonic-moe kernels 

adarshxs:
Adds [SonicMoE](https://github.com/Dao-AILab/sonic-moe) as a Python-only kernel for accelerated Mixture-of-Experts on Hopper and Blackwell GPUs. SonicMoE uses CuTe-DSL grouped GEMMs and Triton routing kernels to deliver state-of-the-art MoE throughput with IO-aware tile scheduling - see [arXiv:2512.14080](https://arxiv.org/abs/2512.14080). The kernel vendors [QuACK](https://github.com/Dao-AILab/quack) (v0.2.5) for its GEMM infrastructure and declares nvidia-cutlass-dsl as a CUDA python-depends. Built and verified with `nix run .#build-and-copy`, and tested end-to-end on H100 via `get_kernel("kernels-community/sonic-moe", version=1)` with forward/backward correctness checks against the torch baseline.


sonic-moe/torch-ext/sonicmoe/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+}
+
+
+def __getattr__(name):
+    if name in _LAZY_IMPORTS:
+        module_path = _LAZY_IMPORTS[name]
+        import importlib
+        mod = importlib.import_module(module_path, __name__)
+        val = getattr(mod, name)
+        globals()[name] = val
```
danieldk: Have you checked if this is compatible with loading different versions of the same kernel (e.g. this not overwriting existing loads)?

This also does a loading attempt on every time a function is accessed.

danieldk: This will probably also be incompatible with `torch.compile`, since it's side-effecting.

sonic-moe/torch-ext/sonicmoe/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+# `import quack` resolves to the bundled copy when no system install exists.
+_vendor_dir = _os.path.join(_os.path.dirname(__file__), "_vendor")
+if _vendor_dir not in _sys.path:
+    _sys.path.insert(0, _vendor_dir)
+
+__version__ = "0.1.1"
+
+# Lazy imports: defer heavy dependencies (cutlass, cuda, triton) so that
+# `import sonicmoe` succeeds in environments without GPU libraries
+# (e.g. the nix build sandbox get-kernel-check).
```
danieldk: What error are you getting in the build sandbox?

sonic-moe/torch-ext/sonicmoe/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+import os as _os
+import sys as _sys
+
+# Inject vendored quack-kernels into the module search path so that
+# `import quack` resolves to the bundled copy when no system install exists.
+_vendor_dir = _os.path.join(_os.path.dirname(__file__), "_vendor")
+if _vendor_dir not in _sys.path:
+    _sys.path.insert(0, _vendor_dir)
+
```
danieldk: This conflicts between different kernels using quack.

Make `quack` a directory under this kernel and use relative imports. See FA4.

sonic-moe/torch-ext/sonicmoe/functional/forward.py
```diff
@@ -0,0 +1,237 @@
...
+from quack.cute_dsl_utils import torch2cute_dtype_map
+
+from ..enums import LIBRARY_NAME, TENSORMAP, ActivationType
+from ..utils import convert_torch_tensor_to_cute_tensor
+from .moe_config import HopperWgmma_MoE_Down_proj_Fwd, HopperWgmma_MoE_Up_proj_Fwd
+from .reduction_over_k_gather import token_gather_and_sum_varlen_K_triton
+from .topk_softmax import TopK_Softmax
+
+
+@torch.library.custom_op(f"{LIBRARY_NAME}::_topk_fwd", mutates_args={"values", "indices"})
```
danieldk: All op registrations need to be prefixed with our namespace (in place of `LIBRARY_NAME`) using `add_op_namespace_prefix`, see e.g.: https://github.com/huggingface/kernels/blob/205a554bb3421a1aa67e941fe94d4418fa8bf0be/examples/kernels/relu-backprop-compile/torch-ext/relu_backprop_compile/__init__.py#L7

Otherwise loading multiple versions will conflict. (The same needs to be done inside quack.

sonic-moe/torch-ext/sonicmoe/__init__.py
adarshxs: it was failing during` get-kernel-check` with `ModuleNotFoundError: No module named 'cutlass'` because the import chain pulled in the cute/quack stack at import time

sonic-moe/torch-ext/sonicmoe/__init__.py
adarshxs: removed. imports are relative now

sonic-moe/torch-ext/sonicmoe/functional/forward.py
adarshxs: correct yes. made the changes

sonic-moe/torch-ext/sonicmoe/__init__.py
adarshxs: yes I verified this by loading two different built revisions. the generated op namespaces differ (_sonicmoe_38e6c2d::... vs _sonicmoe_7f62799::...) and both kernels run side-by-side without collisions.

adarshxs: lazy import path does not break torch.compilen anymore

sonic-moe/torch-ext/sonicmoe/__init__.py
danieldk: That's a bug we should fix in kernel-builder. If `python-depends = ["nvidia-cutlass-dsl"]` is in `build.toml`, it should add `nvidia-cutlass-dsl` to the build sandbox. Oddly enough I don't have this issue in flash-attn4.

https://github.com/huggingface/kernels-community/blob/main/flash-attn4/build.toml

https://github.com/huggingface/kernels-community/blob/main/flash-attn4/torch-ext/flash_attn4/__init__.py

adarshxs: ohkk makes sense. saw your fix

danieldk (approved):
Looks good, great work!

PR #534, opened 2026-04-13
flash-attn2: Add flash_attn_with_kvcache support for XPU

YangKai0616:
Functionality check passed, still need to test the performance.


YangKai0616:
In transformers, PR https://github.com/huggingface/transformers/pull/44379 enables the `flash_attention_with_kvcache` feature to speed up performance for `paged+decode` cases. This PR adds that same feature for XPU, which gives close to a `2x` performance boost.

danieldk (approved):
Looks good! Merged in #842 .

PR #535, opened 2026-04-13
Remove incorrect repo-id check

danieldk:
This check should not be in release builds, plus the command is outdated.


drbh (approved):
lgtm

PR #538, opened 2026-04-14
utilities to sync kernel repos with model repos

sayakpaul:
Until we fully remove the support to load "model" repo types in kernels, we should ensure that the kernel builds are synced for both the repo types ("kernel" and "model") under `kernels-community`. 

This PR adds utilities to do so.

- `scripts/backfill-sync.sh` — One-time script to run locally, iterates over all */build.toml directories (this should be run first, locally)
- `.github/scripts/sync_kernel_to_model.py` — Sync script (Linux/macOS) that downloads from kernel repo type and uploads to model repo type
- `.github/workflows/sync-kernel-model.yaml` — The `workflow_run` orchestrator that watches all three build-release workflows and syncs only after all have completed for the same commit

### DIFF
--- a/.github/scripts/sync_kernel_to_model.py
+++ b/.github/scripts/sync_kernel_to_model.py
@@ -1,11 +1,3 @@
-"""Syncs a kernel repo to its model repo counterpart using huggingface_hub.
-
-Usage: python sync_kernel_to_model.py <repo_id> <branch>
-Example: python sync_kernel_to_model.py kernels-community/flash-attn3 v1
-
-Requires: HF_TOKEN env var, huggingface_hub installed
-"""
-
 import sys
 
 from huggingface_hub import HfApi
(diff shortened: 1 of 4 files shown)


danieldk (commented):
We should update the release workflows to upload both to the kernel and the model repo type. Only two extra lines in each release workflow and no extra script/workflow required.

We only have to retrigger a build for the (I think) three kernels that have been updated since David synced it.

sayakpaul:
There could be a cross-platform race condition problem during the sync no?

> We only have to retrigger a build for the (I think) three kernels that have been updated since David synced it.

For this, should it not be also propagated to the manual build and upload workflow as well, then?

danieldk:
> There could be a cross-platform race condition problem during the sync no?

But there is no sync. Similar to how we are now uploading to both `v1` and `main` branches for `v1` kernels, we'd also upload to the `kernel` repo type. PR: https://github.com/huggingface/kernels-community/pull/539

Seems much simpler than trying to react to the build workflows, etc. and do `gh` output parsing, etc.

sayakpaul:
Closing in favor of https://github.com/huggingface/kernels-community/pull/539

PR #537, opened 2026-04-14
relu: prefer kernel builder cli in ci

drbh:
This PR bumps CI to use the kernel-builder CLI and to push to both kernel and model repo types

PR #539, opened 2026-04-14
relu: upload kernels to kernel repo as well

danieldk:


.github/workflows/build-release-windows.yaml
```diff
@@ -165,7 +165,7 @@ jobs:
         if: steps.validate.outputs.skip == 'false' && steps.check-backend.outputs.supported == 'true'
         working-directory: kernels\kernel-builder
         shell: pwsh
-        run: cargo build
+        run: cargo install
```
danieldk, line 168: So that we can run `kernel-builder` below without additional path. Hope this works on first try, because Windows.

danieldk: I need to update the `repo-type` options after https://github.com/huggingface/kernels/pull/447 is merged.

danieldk:
Closing in favor of https://github.com/huggingface/kernels-community/pull/537

PR #590, opened 2026-04-14
relu: get PR title from env var in `validate-kernel-pr.py`

danieldk:
This avoids having to pass the title as an argument, which opens up the possibility of command injection when not done carefully.


sayakpaul (approved):
Super

PR #591, opened 2026-04-14
relu: only upload to `main` if `version = 1` and no branch is specified

danieldk:


sayakpaul (approved):
Seems secure. Is there chance to inject through any branch name we don't want?

danieldk:
> Seems secure. Is there chance to inject through any branch name we don't want?

I think it's safe indeed, since we are only grepping for the `branch` option, there should be no way to inject anything from the branch name.

PR #662, opened 2026-04-18
scripts: fix kernel existence validation in bot and kernel name display in freshness report

GitGlimpse895:
Two silent bugs in the operations tooling are fixed by this PR.

**1. `pr_comment_kernel_bot.py` — validate kernel exists before dispatching**

When a `/kernel-bot build <kernel>` command is issued with a typo or
a non-existent kernel name, the bot previously dispatched a workflow
that would fail deep inside the Nix build step with no clear signal
to the requester. The bot now checks that the named kernel directory
exists and contains a `build.toml` before dispatching, and posts an
actionable error comment immediately if not.

**2. `check_kernel_freshness.py` — show kernel directory name in Slack report**

The freshness report was formatting stale entries using `source_url`
instead of `kernel_dir`. Since multiple kernels share the same upstream
(e.g. `flash-attn2`, `flash-attn3`, `flash-attn4`, `layer-norm`,
`rotary`, and `vllm-flash-attn3` all point to the same flash-attention
repo), the Slack message would repeat the same URL multiple times with
no indication of which local directory needs updating. Each line now
shows the kernel directory name alongside the upstream URL.

### DIFF
--- a/.github/scripts/pr_comment_kernel_bot.py
+++ b/.github/scripts/pr_comment_kernel_bot.py
@@ -643,7 +643,26 @@ def main():
     dispatches = []
     failed = []
 
+    repo_root = os.path.dirname(os.path.dirname(os.path.abspath(__file__)))
+
     for kernel_name in kernels:
+        kernel_path = os.path.join(repo_root, kernel_name)
+        if not os.path.isdir(kernel_path) or not os.path.isfile(os.path.join(kernel_path, "build.toml")):
+            try_send_issue_comment(
+                api_base,
+                token,
+                issue_number,
+                format_result_comment(
+                    command_summary,
+                    mode_text,
+                    target_branch,
+                    pr_head_sha,
+                    failure_message=f"Kernel `{kernel_name}` was not found in this repository. Check the spelling and try again.",
+                ),
+                comment_id=status_comment_id,
+            )
+            return 0
+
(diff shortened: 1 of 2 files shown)


GitGlimpse895:
Hi @danieldk and @drbh — when you get a moment, would you mind taking a look at this PR? It's a small, focused fix across two scripts:

- `pr_comment_kernel_bot.py` now validates that a kernel directory and its `build.toml` exist before dispatching, so typos surface immediately as a clear comment rather than failing silently mid-build.
- `check_kernel_freshness.py` now shows the kernel directory name in the Slack report instead of the source URL, which makes stale entries much easier to act on when multiple kernels share the same upstream (e.g. the flash-attention family).

No logic changes to the build or freshness-check logic itself — just validation and display improvements. Happy to address any feedback. Thanks so much for your time! 🙏

drbh:
Thanks for putting this together - the improvements make sense. However, I don't think we are planning to merge these changes. This tooling is internal and still in flux, and we’d prefer not to add incremental fixes while it’s likely to change more substantially.

Appreciate you taking the time to put this together. Additionally we are exploring a new contribution workflow and soon are going to require that users open issues, before opening PRs. Please feel free to open any issues for other bugs that you'd like to flag. 

Thanks!

PR #667, opened 2026-04-21
sonic-moe: update to support new quack and cutlass

adarshxs:


sonic-moe/torch-ext/sonic_moe/__init__.py
```diff
@@ -2,23 +2,15 @@
 # Copyright (c) 2025, Wentao Guo, Mayank Mishra, Xinle Cheng, Ion Stoica, Tri Dao
 # ********************************************************************************
 
-from functools import lru_cache
-
-__version__ = "0.1.1"
+__version__ = "0.1.2.post1"
```
IlyasMoutawwakil, line 5: is this the right version ?

danieldk (approved):

PR #674, opened 2026-04-28
gpt-oss-triton-kernels: Fix backend check for ROCm targets.

Exile333:
Nearly duplicate of https://huggingface.co/kernels-community/gpt-oss-triton-kernels/discussions/2

This fix is necessary to run gpt-oss model on AMD hardware. When running this model at AMD GPUs, `is_cuda()` will return `False`, and `is_hip()` is never called. This causes function to return None, which causes exceptions on caller's side. Pytorch uses cuda submodule when is built for HIP backend, so we want to use it at HIP backend.

### DIFF
--- a/gpt-oss-triton-kernels/torch-ext/gpt_oss_triton_kernels/target_info.py
+++ b/gpt-oss-triton-kernels/torch-ext/gpt_oss_triton_kernels/target_info.py
@@ -92,7 +92,7 @@ def has_native_mxfp():
 
 
 def num_sms():
-    if is_cuda():
+    if is_cuda() or is_hip():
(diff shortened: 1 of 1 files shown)


Exile333:
@danieldk @drbh 
Please, take a look when you have time.
Feel free to redirect me, if this is not a right place for patches.

drbh:
Hey @Exile333 thanks for the contribution, I just reopened these changes in https://github.com/huggingface/kernels-community/pull/679 to run CI. once it finishes I'll approve and merge this PR

*also made a small update to the PR title to trigger a rebuild of the gpt-oss-triton-kernels kernel

drbh (approved):
lgtm

PR #679, opened 2026-04-29
[CI] Fix backend check for ROCm targets.

drbh:
Nearly duplicate of https://huggingface.co/kernels-community/gpt-oss-triton-kernels/discussions/2

This fix is necessary to run gpt-oss model on AMD hardware. When running this model at AMD GPUs, `is_cuda()` will return `False`, and `is_hip()` is never called. This causes function to return None, which causes exceptions on caller's side. Pytorch uses cuda submodule when is built for HIP backend.


drbh:
closing as https://github.com/huggingface/kernels-community/pull/674 was merged

PR #680, opened 2026-04-30
finegrained-fp8: [bugfix] add epsilon to act_quant scale to avoid NaNs.

### LINKED ISSUE #681: Bug: finegrained-fp8 act_quant produces NaNs on zero blocks
Heya, quick note that finegrained-fp8's [act_quant kernel](https://github.com/huggingface/kernels-community/blob/main/finegrained-fp8/torch-ext/finegrained_fp8/act_quant.py) computes the per-block FP8 scale with `s = max(abs(x)) / 448` with no eps, so we get `s = 0` when x is a block of zeros, and the 0/0 division yields NaN. 

I opened a PR for it earlier but didn't realize that it was contributor-only: https://github.com/huggingface/kernels-community/pull/680

wz-ml:
A quick single-line change to fix this issue I was running into with this FP8 kernel:
When an X block is all zeros, the scale factor's also zero, which causes the kernel to return 0/0 = NaN.


danieldk:
@IlyasMoutawwakil could you do a review?

finegrained-fp8/torch-ext/finegrained_fp8/act_quant.py
```diff
@@ -30,6 +30,7 @@ def _fp8_act_quant_kernel(x_ptr, y_ptr, s_ptr, BLOCK_SIZE: tl.constexpr):
     offs = pid * BLOCK_SIZE + tl.arange(0, BLOCK_SIZE)
     x = tl.load(x_ptr + offs).to(tl.float32)
     s = tl.max(tl.abs(x)) / 448.0  # float8_e4m3fn max
+    s = tl.maximum(s, 1e-12)
```
IlyasMoutawwakil, line 33: yes this actually needed, thanks for the fix.

IlyasMoutawwakil:
it seems i can't approve the pr 🥲

sayakpaul:
@IlyasMoutawwakil this looks good to you?

IlyasMoutawwakil (approved):

IlyasMoutawwakil:
@sayakpaul yes i couldn't approve it before

PR #823, opened 2026-05-05
deep-gemm: Port upstream DeepGEMM updates

adarshxs:
Port recent upstream DeepGEMM sources and Python APIs into `deep-gemm`, update Kernel Hub build metadata for supported CUDA toolchains/architectures, and add local wrapper compatibility fixes for PyTorch limited ABI and DeepGEMM runtime JIT init.


drbh (dismissed):
lgtm

deep-gemm/csrc/jit_kernels/impls/runtime_utils.hpp
```diff
@@ -64,8 +66,12 @@ static std::string to_string(const at::ScalarType& dtype) {
...
     if (std::isfinite(v)) {
         char buffer[32];
-        std::snprintf(buffer, sizeof(buffer), "%.9gf", v);
-        return buffer;
+        std::snprintf(buffer, sizeof(buffer), "%.9g", v);
+        std::string literal = buffer;
+        if (literal.find_first_of(".eE") == std::string::npos)
+            literal += ".0";
+        literal += "f";
+        return literal;
```
danieldk: Why is this changed compared to upstream? This is incorrect, since `snprintf` is locale-dependent, so it will return incorrect string literals for people that have their locale set to one that uses `,` for decimal separators.

In general, we want to stay absolutely as close as possible to upstream. When these subtle breakages happen, people will report them upstream and we do not want to put any extra burden on upstreams.

adarshxs: yes agree and apologies for the divergence but for context, this was added after @IlyasMoutawwakil hit a runtime JIT failure in the MegaMoE path: with V4 flash type `swiglu_limit=10.0`, the generated C++ literal became `10f`, which NVCC rejected as a user-defined literal instead of treating it as a float. 

I think @IlyasMoutawwakil had a temporary python integration side workaround that worked? but this was an attempt to fix at source. @IlyasMoutawwakil could you suggest a way to tackle this?

IlyasMoutawwakil: my fix wasn't really a fix, i just added an epsilon to 10.0 to force it be 10.000001 before calling megamoe.

IlyasMoutawwakil (dismissed):
LGTM ! awesome work !

deep-gemm/build.toml
```diff
@@ -77,32 +81,58 @@ src = [
...
+  "deep_gemm/include/deep_gemm/ptx/utils.cuh",
+  "deep_gemm/include/deep_gemm/ptx/wgmma.cuh",
+  "deep_gemm/include/deep_gemm/scheduler/gemm.cuh",
+  "deep_gemm/include/deep_gemm/scheduler/mega_moe.cuh",
+  "deep_gemm/include/deep_gemm/scheduler/paged_mqa_logits.cuh",
 ]
 
 [general.hub]
 repo-id = "kernels-community/deep-gemm"
+branch = "v2"
```
IlyasMoutawwakil: this is not needed anymore right ?
```suggestion
```

deep-gemm/build.toml
```diff
@@ -16,46 +16,49 @@ pyext = ["py", "cuh", "hpp", "h"]
...
 cuda-flags = [
-  "-std=c++17",
+  "-std=c++20",
   "-O3",
   "-DNDEBUG",
   "--expt-relaxed-constexpr",
   "--expt-extended-lambda",
   "-Wno-deprecated-declarations",
 ]
 src = [
```
IlyasMoutawwakil, line 34:
```suggestion
src = [
  # Compiled source
```

deep-gemm/build.toml
```diff
@@ -34,28 +34,32 @@ cuda-flags = [
...
   # Compiled source
   "csrc/impl.cu",
-  # API headers
   "csrc/apis/attention.hpp",
```
IlyasMoutawwakil, line 38:
```suggestion
 # API headers
  "csrc/apis/attention.hpp",
```

danieldk:
Merged in #895, thanks!

PR #826, opened 2026-05-07
relu: spin off separate CI jobs for different backends on x86_64-linux

danieldk:


drbh (approved):
nice, excited to see how this impacts CI time

PR #839, opened 2026-05-11
Fix duplicate-address store race in batched FP8 matmul kernels (XPU c…

sywangyi:
…orrectness)

Summary
w8a8_block_fp8_matmul_batched_kernel and w8a8_tensor_fp8_matmul_batched_kernel issue BLOCK_SIZE_M-way duplicate-address tl.stores to the output, which is undefined behavior. On NVIDIA GPUs the writes happen to be benign (last-write-wins of byte-identical values), but on Intel XPU it corrupts the output. This PR adds a single-lane store mask to make the write well-defined.

Fix:
Mask the store so only lane 0 writes:

### DIFF
--- a/finegrained-fp8/flake.lock
+++ b/finegrained-fp8/flake.lock
@@ -41,16 +41,15 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
-        "lastModified": 1777901020,
-        "narHash": "sha256-m54hEdh1kK+pDcRfJDwN0P861X5VWpC/yZ1TjTwshtA=",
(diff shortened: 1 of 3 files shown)


sywangyi:
find the error during enabling deepseek v4 flash base modeling in xpu

sywangyi:
@danieldk pls help review

IlyasMoutawwakil (approved):
lgtm !

danieldk:
Thanks a lot! Merged in https://github.com/huggingface/kernels-community/pull/846.

PR #845, opened 2026-05-13
deep-gemm: improve security

sayakpaul:
We had a security reporting:

"""
[LOW] .github/workflows/security-audit.yml:112-113 — Audit prompt drops explicit coverage of setup.py, pyproject.toml, and CMake ExternalProject_Add as fetch-at-build-time vectors

The removed text told the LLM auditor to specifically check "URLs in setup.py, pyproject.toml, CMake ExternalProject_Add" for unpinned network fetches. This is directly relevant: deep-gemm/setup.py:51,168-191 contains a CachedWheelsCommand that downloads pre-built wheels from https://github.com/DeepSeek-AI/DeepGEMM/releases/ at build time via urllib.request.urlopen with no SHA verification — only a version-based URL. An attacker who modifies the base_wheel_url or poisons the GitHub release would serve a malicious wheel to every builder. Removing setup.py from the explicit checklist makes it less likely the auditor will flag changes to this download path. The general principle ("Anything that pulls bytes from the network at build or import time without a SHA pin") still applies, but LLM auditors are measurably more reliable when given specific file types to check.
"""

This PR fixes that.


IlyasMoutawwakil (approved):
lgtm !

vasqu (approved):
Thank you, makes sense!

PR #857, opened 2026-05-15
liger-kernels: sync with liger-kernels upstream

sayakpaul:
Also update `AGENTS.md` to include Liger-Kernels specific instructions for syncing with the upstream.

Cc: @vasqu


AGENTS.md
```diff
@@ -35,3 +35,44 @@ steps:
 
 If the user did not specify the version tag, stop and ask which tag to sync
 from.
+
+## liger-kernels
```
danieldk, line 39: Nice!!!

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -1,39 +1,345 @@
...
+            self.reduction,
+            self.softcap,
+            False,
+            False,
+            False,
+        )
+        return loss
+
+
+class LigerFusedLinearCrossEntropyLoss(nn.Module):
```
vasqu, line 182: Could we also expose something like https://github.com/linkedin/Liger-Kernel/blob/38c0d4f8af6cc0386a4c798558e992570843f1ac/src/liger_kernel/transformers/model/loss_utils.py#L67 in the init? That's at least exactly what we are looking for in transformers then

I.e. we are looking into the base function we will exchange instead of the module

sayakpaul: Check now?

sayakpaul:
@vasqu the build CI is here https://github.com/huggingface/kernels-community/actions/runs/26037106273/job/76538369834

liger-kernels/torch-ext/liger_kernels/layers.py
vasqu: Yep, looking good now, thanks!

vasqu (approved):

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -1,39 +1,463 @@
...
+    def __init__(
+        self,
+        hidden_size: int,
+        eps: float = 1e-6,
+        offset: float = 0.0,
+        casting_mode: str = "llama",
+        init_fn: str = "ones",
+        in_place: bool = True,
+        row_mode: Optional[bool] = None,
+        elementwise_affine: bool = True,
```
vasqu, line 33: I think here we can ignore the extra init args and always default as done before?

Edit: Don't think it matters too much tho, should be fine as is as well

sayakpaul: Cool

PR #864, opened 2026-05-17
fix: correct return type annotations from None to torch.Tensor in activation module

### LINKED ISSUE #848: activation: return type annotations incorrectly declare `-> None` instead of `-> torch.Tensor`
## Bug Description

All 11 public functions in `activation/torch-ext/activation/__init__.py`
are annotated with `-> None` but every one of them actually returns `out`
(a `torch.Tensor`) via an explicit `return out` statement.

## Affected Functions

`silu_and_mul`, `mul_and_silu`, `gelu_and_mul`, `gelu_tanh_and_mul`,
`fatrelu_and_mul`, `gelu`, `silu`, `gelu_tanh`, `gelu_fast`, `gelu_new`,
`gelu_quick`

## Impact

- **Type checkers** (mypy / pyright) raise errors at any call site that
  stores the return value — including the project's own test at
  `activation/tests/kernels/test_activation.py:117`:
  `out = fn(out, x)`
- **Misleading API contract**: callers reading the signature would assume
  the functions are purely in-place and discard the return value
- **Incorrect auto-generated docs / stubs** for downstream integrators

## Minimal Reproduction

```python
import inspect
import activation

# All of these will show -> None despite returning torch.Tensor
print(inspect.signature(activation.silu_and_mul))  # (out: Tensor, x: Tensor) -> None
```

## Fix

Change `-> None` to `-> torch.Tensor` on all 11 signatures.

**Before:**
```python
def silu_and_mul(out: torch.Tensor, x: torch.Tensor) -> None:
    ops.silu_and_mul(out, x)
    return out
```

**After:**
```python
def silu_and_mul(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
    ops.silu_and_mul(out, x)
    return out
```

No logic or behaviour changes — purely an annotation correction.
The `return out` statements already exist in every function.

## File Reference

[`activation/torch-ext/activation/__init__.py`](https://github.com/huggingface/kernels-community/blob/main/activation/torch-ext/activation/__init__.py)

Nitin-Rajasekar:
Fixes #848

All 11 public functions in `activation/torch-ext/activation/__init__.py` were annotated with `-> None` despite explicitly returning `out` (a `torch.Tensor`). This corrects all 11 signatures to `-> torch.Tensor`.

No logic or behaviour changes, purely an annotation fix.

### DIFF
--- a/activation/torch-ext/activation/__init__.py
+++ b/activation/torch-ext/activation/__init__.py
@@ -5,56 +5,56 @@ from ._ops import ops
 from . import layers
 
 
-def silu_and_mul(out: torch.Tensor, x: torch.Tensor) -> None:
+def silu_and_mul(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
     ops.silu_and_mul(out, x)
     return out
 
 
-def mul_and_silu(out: torch.Tensor, x: torch.Tensor) -> None:
+def mul_and_silu(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
     ops.mul_and_silu(out, x)
     return out
 
 
-def gelu_and_mul(out: torch.Tensor, x: torch.Tensor) -> None:
+def gelu_and_mul(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
     ops.gelu_and_mul(out, x)
     return out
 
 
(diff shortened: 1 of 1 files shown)


Nitin-Rajasekar:
Hi @danieldk, looks like this project requires vouched users..Could you let me know how I can get vouched so this fix can be considered (or if that it possible for me at all)? Happy to do whatever is needed.

PR #867, opened 2026-05-18
finegrained-fp8: add mxfp4 recipes

sywangyi:
verified in xpu and b200.
- fp4 (mxfp4 style) with inline quantization.


sywangyi:
@IlyasMoutawwakil please help review

IlyasMoutawwakil:
shouldn't they be w4a8 ?

sywangyi:
no, activation is 16bit since fp8_act_quant is not called in activation in fp4 kernels path

sywangyi:
To do FP8 × FP4, the activations cannot be used as-is. They must first be quantized, typically per-token or per-block, together with the corresponding scale factors. In the MoE grouped path, there is already additional overhead from routing, sorting, the activation function, and the second projection, so activation quantization, dequantization, and scale movement are not free. FP8 activations only provide a clear speedup when the backend fuses activation quantization and GEMM efficiently. DeepGEMM does that; this Triton fallback does not.

IlyasMoutawwakil:
yes exactly and that's why we do it in the batched and grouped fp8 paths as well so why not in these fp4 ones 😅 for me this is not much of a choice but rather how to stay as close to the original dsv4 implementation.

finegrained-fp8/torch-ext/finegrained_fp8/batched.py
```diff
@@ -380,16 +381,20 @@ def w8a8_fp8_matmul_batched(
     expert_ids: torch.Tensor,
     block_size: list[int] | None,
 ) -> torch.Tensor:
-    """Unified batched W8A8 FP8 matmul dispatcher.
+    """Unified batched W8A8/FP4 matmul dispatcher.
```
IlyasMoutawwakil: i don't think it makes sense to add them here, the name of the functions has "w8a8_fp8"

sywangyi: I agree the naming is not ideal if these were meant to be new public FP4-specific APIs.

My intention here was the opposite: to avoid introducing a separate FP4 API surface in transformers and keep the transformers-side change minimal. Today transformers already calls the w8a8_fp8_* entrypoints, so routing the FP4-weight path through the same dispatchers lets the integration stay almost unchanged and keeps the FP4 handling internal to kernels-community.

IlyasMoutawwakil: here's an approach that's both semantically correct and bc ; we can introduce another dispatcher called fp8_batched/grouped_matmul which dispatches between w8a8 and w4a8. And then in a transformers PR we start using that instead of w8a8. Transformers moves forward, the kernel on the hub stays bc and semantically correct.

sywangyi: I’m happy to switch to that approach: add the new fp8_matmul / fp8_batched_matmul / fp8_grouped_matmul dispatchers here, keep the existing w8a8_fp8_* paths unchanged for BC, and then update transformers to call the new dispatcher in a follow-up PR.

sywangyi:
test in Intel GPU 1550 using benchmark_perf.py
acceleration ratio 
[matmul] deepseek_v4_gate_up_decode | M=32 N=4096 K=4096 block_n=128 block_k=128 | median=0.5170ms mean=0.5170ms min=0.5170ms max=0.5170ms effective=2.077 TOPS
  baseline=torch_fp4_baseline median=1.4965ms speedup=2.894x
[batched] deepseek_v4_gate_up_decode | S=192 E=256 N=4096 K=4096 top_k=6 block_n=128 block_k=128 | median=13.0566ms mean=13.0566ms min=13.0566ms max=13.0566ms effective=0.493 TOPS
  baseline=torch_fp4_baseline median=195.1542ms speedup=14.947x
[grouped] deepseek_v4_gate_up_decode | S=192 E=256 N=4096 K=4096 top_k=6 block_n=128 block_k=128 | median=11.4190ms mean=11.4190ms min=11.4190ms max=11.4190ms effective=0.564 TOPS
  baseline=torch_fp4_baseline median=195.8510ms speedup=17.151x

[matmul] deepseek_v4_gate_up_prefill | M=256 N=4096 K=4096 block_n=128 block_k=128 | median=1.3954ms mean=1.3954ms min=1.3954ms max=1.3954ms effective=6.156 TOPS
  baseline=torch_fp4_baseline median=1.8095ms speedup=1.297x
[batched] deepseek_v4_gate_up_prefill | S=1536 E=256 N=4096 K=4096 top_k=6 block_n=128 block_k=128 | median=96.9130ms mean=96.9130ms min=96.9130ms max=96.9130ms effective=0.532 TOPS
  baseline=torch_fp4_baseline median=485.1562ms speedup=5.006x
[grouped] deepseek_v4_gate_up_prefill | S=1536 E=256 N=4096 K=4096 top_k=6 block_n=128 block_k=128 | median=21.8489ms mean=21.8489ms min=21.8489ms max=21.8489ms effective=2.359 TOPS
  baseline=torch_fp4_baseline median=485.0706ms speedup=22.201x

IlyasMoutawwakil (dismissed):
LGTM !

IlyasMoutawwakil (dismissed):

IlyasMoutawwakil (approved):

PR #874, opened 2026-05-19
feat: report build outcome as comment on PR that triggered build

drbh:
This PR adds a step to report the build outcome to the PR that trigged the the build.


drbh:
optimistically merging as this adds a small step that write a comment on the PR that triggered the build and needs to be merged to test. will revert if any issues

PR #877, opened 2026-05-19
flash-attn2: prefer using add_op_namespace_prefix

drbh:
This PR fixes flash-attn2 to correctly register fake ops


sayakpaul (dismissed):
Thanks!

drbh:
note this PR has been updated to remove the unused ops folder that included non flash attn code that was unused. It also cleans up the exposed functions to removed the unused low levels functions in __init__. 

the kernel now only exposes the core top level functions (listed in __all__) and passes the `python kernels/nix-builder/pkgs/torch-ops-check/torch-ops-check-hook.py kernels-community/flash-attn2/torch-ext` check add in https://github.com/huggingface/kernels/pull/569

sayakpaul (commented):
Any reason why the windows build would fail?

drbh:
> Any reason why the windows build would fail?

not 100% sure at the moment but seems to be related to the xpu path on windows. In general the windows build workflow may need some tweaks since it has some custom logic that diverges from the standard kernel-builder nix path. 

gonna take a look and see if there is a small change to resolve - otherwise fixing the workflow may be best to tackle in another PR

sayakpaul:
Works for me.

drbh:
added a small PR to skip the windows xpu backend for flash attn2 since there seems to be a bug related to the cutlass fork, its possible that merging that PR and rebasing this PR on top will avoid the xpu windows build and enable the windows cuda build to succeed.. https://github.com/huggingface/kernels-community/pull/885

danieldk (dismissed):

flash-attn2/torch-ext/flash_attn2/__init__.py
```diff
@@ -11,383 +8,12 @@
     flash_attn_with_kvcache,
 )
 
-
-def fwd(
```
danieldk, line 15: This looks like an API break, we need to bump up the version if we remove these. Is removal necessary?

vasqu:
Based on #894, we likely also need to sync with the latest stable version `2.8.3` instead of `main` - the issue(s) are described there

flash-attn2/torch-ext/flash_attn2/__init__.py
drbh: good point, Ive added the low level functions back and exposed them in __all__

danieldk (approved):

PR #880, opened 2026-05-19
[ci] don't let vouch auto-close a pr re-opened by a maintainer.

sayakpaul:


.github/workflows/vouch-check-pr.yaml
```diff
@@ -12,7 +12,25 @@ jobs:
     name: Check PR author
     runs-on: ubuntu-latest
     steps:
-      - uses: mitchellh/vouch/action/check-pr@c6d80ead49839655b61b422700b7a3bc9d0804a9 # v1.4.2
+      - name: Check if reopened by a maintainer
+        id: maintainer
```
danieldk, line 16: Question, currently we use:

```
  pull_request_target:
    types: [opened, reopened]
```

Can an unprivileged user re-open a PR? If not, I think we can achieve the same by removing `reopened` above?

sayakpaul: Good point, but I think the PR author themselves can reopen their own PR even without write access (per GitHub's permission model — anyone who opens a PR can close/reopen it). So if we drop reopened, an unvouched author could reopen the auto-closed PR and it would stay open without re-triggering the check.

That said, if that's an acceptable trade-off (closing once on opened is enough as a signal), removing reopened would indeed be the simpler fix.

LMK which route you would prefer.

danieldk (approved):

.github/workflows/vouch-check-pr.yaml
danieldk: Ah, that's annoying, let's do it as in your PR then.

PR #885, opened 2026-05-20
feat: allow skipping backends on windows ci

drbh:
This PR allows specific kernels to skip listed backends on the windows ci build workflow. Namely this skips the windows xpu build for flash-attn2 since there is a bug with a type in the cutlass fork https://github.com/huggingface/kernels-community/actions/runs/26142413949/job/76890482527#step:17:639

### DIFF
--- a/.github/scripts/dispatch.py
+++ b/.github/scripts/dispatch.py
@@ -125,6 +125,11 @@ WINDOWS_KERNELS = {
     "flash-attn2",
 }
 
+# Backends to skip on Windows for specific kernels (e.g. due to toolchain issues).
+WINDOWS_SKIP_BACKENDS: dict[str, set[str]] = {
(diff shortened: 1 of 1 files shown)


.github/scripts/dispatch.py
```diff
@@ -125,6 +125,11 @@ def get_repo() -> str | None:
     "flash-attn2",
 }
 
+# Backends to skip on Windows for specific kernels (e.g. due to toolchain issues).
+WINDOWS_SKIP_BACKENDS: dict[str, set[str]] = {
+    "flash-attn2": {"xpu"},  # CUTLASS XPU headers fail on Windows (ushort undefined)
```
sayakpaul, line 130: Is is something we should fix syncing with upstream?

sayakpaul (approved):

danieldk (approved):

PR #899, opened 2026-05-29
flash-attn3: sync with upstream and switch to the Torch stable ABI

danieldk:
Also add agent sync instructions.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,5 +1,43 @@
 # Kernel-specific instructions
 
+## flash-attn3
+
+When the user asks to sync a flash-attn3 release, carry out the following
+steps:
+
+- Fetch the upstream Git repository from https://github.com/Dao-AILab/flash-attention.git
+- Check out the tag that the user specified.
+- Flash Attention 3 is in the directory `hopper` of the upstream repo.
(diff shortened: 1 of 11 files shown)


AGENTS.md
```diff
@@ -1,5 +1,43 @@
 # Kernel-specific instructions
 
+## flash-attn3
+
+When the user asks to sync a flash-attn3 release, carry out the following
+steps:
+
+- Fetch the upstream Git repository from https://github.com/Dao-AILab/flash-attention.git
+- Check out the tag that the user specified.
```
sayakpaul, line 9: This is what got me confused before. Should there be instructions on fetching a stable tag? I guess FA3 is beta but this should be the case for kernels that have stable releases?

Not for this PR, though.

sayakpaul (approved):

AGENTS.md
danieldk: For stable kernels we could do something like _Check out the tag that corresponds to the latest stable version_? Not sure how well that works, maybe if we also point it to a GitHub release page.

sayakpaul: We could also have a Python utility that fetches the latest tag and I think that way, we will have a more definitive solution.

PR #906, opened 2026-06-01
liger-kernels: sync changes for transformers

vasqu:
As per title, I need a few changes for transformers to work properly
- Signature fixes
- Init fixes
- Torch compile fixes

Also dependent on https://github.com/huggingface/kernels/pull/607 as this enables functions with torch compile (but it won't break without the other PR either way, it would only unlock the feature)

NOTE: This assumes full transformers usage and only adjusts the modules that we need, nothing else


liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -20,49 +20,24 @@
 from .tvd import LigerTVDLossFunction
 
 
+# NOTE: Not compile-friendly --> large deviations to the original implementation under compile
 class LigerRMSNorm(nn.Module):
-    def __init__(
```
vasqu, line 25: No init as we need stateless modules

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -358,23 +325,32 @@ def liger_fused_linear_cross_entropy(
     )
 
 
+# NOTE: We have this intentional graph break as we encounter issues such as IMAs and Cublas errors.
+#       We know that this is an optimized kernel already so there is less ways to
+#       fuse it either way; we rely on torch compile to go through the base model to optimize.
+@torch._dynamo.disable
```
vasqu: This is an intentional graph break, I played a lot around training and I came to the decision that this is the best as we still keep the benefit of the kernel and kernel fusions with torch compile would be not really possible either way

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -443,6 +404,11 @@ def liger_multimodal_rotary_pos_emb(
     return LigerQwen2VLMRopeFunction.apply(q, k, cos, sin, mrope_section, unsqueeze_dim)
 
 
+# Add torch compile support for functions - in this case it's to allow this to be used
+# but the function itself will not be compiled (see the note at the function)
+LigerForCausalLMLoss.can_torch_compile = True
```
vasqu, line 252: This is what I meant re the other PR to enable torch compile for functions

sayakpaul, line 252: It can go within the layer init like `LigerSwiGLUMLP`?

liger-kernels/torch-ext/liger_kernels/layers.py
sayakpaul: (nit): Would prefer using public APIs: `torch.compiler.disable`.

Agree on the reasoning:

> We know that this is an optimized kernel already so there is less ways to
 fuse it either way; we rely on torch compile to go through the base model to optimize.

Some additional thoughts.

My impression was that if we register the op with `@torch.library.custom_op()`, it would include the ops during compilation and wouldn't try tracing through it. So, maybe we would like to add that as a decorator to the underlying kernel functions `liger_fused_linear_cross_entropy`?

And then on the `transformers` side, there could be a general utility like `maybe_disable_torch_compile` for cases like this. IMO, we probably shouldn't add disable decorators ourselves and let users handle it. But if we go the route of maintaining the "transformers-specific" source, then that's a different story.

vasqu: Hmm, yea let me try if I could do custom ops wrappers. Definitely better long term but maybe that should live in the original liger kenerl repo. I was not sure how triton was treated during compile but ig it's not as I wanted

Will split the implementations into transformers x original based on the suggestion no worries

liger-kernels/torch-ext/liger_kernels/layers.py
vasqu: It's a function there is no init so we kinda have to attach it afterwards no?

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -20,49 +20,24 @@
 from .tvd import LigerTVDLossFunction
 
 
+# NOTE: Not compile-friendly --> large deviations to the original implementation under compile
 class LigerRMSNorm(nn.Module):
```
danieldk, line 17: Could you bump up the version in `build.toml` as well, since it's an API change?

liger-kernels/torch-ext/liger_kernels/layers.py
danieldk: So the motivation of adding 'function layers' is that in transformers, some things are currently functions and then it's more logical to extend using functions.

That said, I think I made it work both ways, so you can also extend a function using a (stateless) layer.

danieldk:
Note that until we fully switch to the bot, the PR title must have the format

```
<kernel name>: description
```

Also see the CI result:

```
No colon found in PR title, skipping build
```

vasqu:
Sorry guys I was delayed by the gemma release, I will try to get to this today:
- Reverting to old state and adding transformers specific interfaces --> best of both worlds
- Try to take a look into wrapping into fake ops, although might be better as an upstream fix. Will see what I can do

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -358,23 +93,32 @@ def liger_fused_linear_cross_entropy(
     )
 
 
+# NOTE: We have this intentional graph break as we encounter issues such as IMAs and Cublas errors.
+#       We know that this is an optimized kernel already so there is less ways to
+#       fuse it either way; we rely on torch compile to go through the base model to optimize.
+@torch.compiler.disable
```
vasqu, line 195: I would like to keep this for now. If we were to change this with fake ops etc, it would be something that would need to be done upstream tbh because it touches the core layers

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -1,307 +1,84 @@
...
-        self.variance_epsilon = eps
-        self.weight = nn.Parameter(torch.ones(hidden_size) if init_fn == "ones" else torch.zeros(hidden_size))
-        self.bias = nn.Parameter(torch.randn(hidden_size) if bias else torch.zeros(hidden_size))
+    weight: nn.Parameter
+    bias: nn.Parameter | None
+    eps: float
 
-    def forward(self, hidden_states: torch.Tensor) -> torch.Tensor:
-        return LigerLayerNormFunction.apply(hidden_states, self.weight, self.bias, self.variance_epsilon)
+    def forward(self, input: torch.Tensor) -> torch.Tensor:
```
vasqu: We use torch's native nn.LayerNorm so we need to be strict here

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -1,307 +1,84 @@
...
+def liger_rotary_pos_emb(
+    q: torch.Tensor,
+    k: torch.Tensor,
+    cos: torch.Tensor,
+    sin: torch.Tensor,
+    position_ids: Optional[torch.Tensor] = None,
+    unsqueeze_dim: int = 1,
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Apply standard rotary positional embedding to ``q`` and ``k``."""
+    return LigerRopeFunction.apply(q, k, cos, sin, position_ids, unsqueeze_dim)
```
vasqu: Honestly we could also use this instead of rotary or FA one's - these will be at least properly maintained cc @sayakpaul

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -311,7 +88,8 @@ class LigerTiledGEGLUMLP(nn.Module):
     gate_proj: nn.Linear
     up_proj: nn.Linear
     down_proj: nn.Linear
-    num_shards: int
```
vasqu, line 314: If we want to use these as opt in then we don't have this attribute in transformers but let me double check if things work as expected. This could be the training version

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -435,80 +223,36 @@ def LigerForCausalLMLoss(
...
 __all__ = [
-    "LigerRMSNorm",
     "LigerLayerNorm",
-    "LigerGroupNorm",
-    "LigerDyT",
-    "LigerCrossEntropyLoss",
-    "LigerFusedLinearCrossEntropyLoss",
-    "LigerJSD",
-    "LigerKLDIVLoss",
-    "LigerTVDLoss",
```
vasqu, line 504: I deleted a bunch because their exact usage is not too obv imo especially since a lot of these modules should rather be functions (losses). We can always get another version or progressively update/add(?)

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -1,307 +1,104 @@
...
-class LigerDyT(nn.Module):
-    def __init__(self, hidden_size: int, beta: bool = True, init_alpha: float = 0.5):
-        super().__init__()
-        self.hidden_size = hidden_size
-        self.init_alpha = init_alpha
-        self.alpha = nn.Parameter(torch.ones(1) * init_alpha)
-        self.gamma = nn.Parameter(torch.ones(hidden_size))
-        self.beta = nn.Parameter(torch.zeros(hidden_size)) if beta else None
+# TODO: experimental --> same as tiled but generalized to one MLP
+class LigerMLP(nn.Module):
```
vasqu: Experimental but this could be very nice in general - might delete but want to keep this here for now

vasqu:
Closing in favor of #948 to have a non-fork branch to be used (seems like some gh actions are dependent on write access so they fail here)

PR #907, opened 2026-06-01
rotary: Sync changes for transformers

vasqu:
Similar in fashion to #906 but we only need the compile flag tbh


danieldk (approved):

vasqu:
cc @danieldk would you mind doing the command :face_holding_back_tears:

PR #912, opened 2026-06-02
aiter-flash-attn: ship MHA configs in Hub artifact

Abdennacer-Badaoui:
Ship MHA configs in Hub artifact

### DIFF
--- a/aiter-flash-attn/build.toml
+++ b/aiter-flash-attn/build.toml
@@ -9,6 +9,7 @@ backends = [
 [general.hub]
 repo-id = "kernels-community/aiter-flash-attn"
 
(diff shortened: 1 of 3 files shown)


aiter-flash-attn/build.toml
```diff
@@ -9,6 +9,7 @@ backends = [
 [general.hub]
 repo-id = "kernels-community/aiter-flash-attn"
 
-[torch-noarch]
+[torch]
```
danieldk: Should not be used for noarch kernel. Discussing a fix in Slack.

danieldk (approved):

PR #916, opened 2026-06-02
megablocks: add ROCm backend (AITER Triton grouped GEMM)

Abdennacer-Badaoui:
Adds AMD ROCm support to the megablocks kernel alongside the existing cuda/xpu/cpu backends.

  - New `[kernel.megablocks_rocm]` block (gfx942 + gfx950) reusing the same CUDA op sources via hipify; two small CUDA-safe guards (`_LDG` for `__ldg`, `MEGABLOCKS_CUB` for `cub::DeviceHistogram`) plug the gaps in torch's hipify symbol table.
  - CUTLASS grouped GEMM is CUDA-only; on ROCm grouped GEMM is provided by AITER's Triton `gmm`/`ptgmm`, vendored under `torch-ext/megablocks/_grouped_gemm_triton/` with an adapter mapping megablocks' `gmm(a, b, c, batch_sizes, trans_a, trans_b)` semantics. `grouped_gemm/backend.py` dispatches by `torch.version.hip`.
  - `torch_binding.cpp` registers the 5 ops under `ROCM_KERNEL` (HIP → `torch::kCUDA`); CUTLASS `gmm` stays CUDA-only.
  - Drive-by: fixes a latent shadowing bug in `megablocks/__init__.py` where `from . import layers` rebinds `ops` to the python submodule, breaking the top-level `megablocks.sort` wrapper on both backends.
  
  Validated on gfx942 (MI300X): ops correctness, grouped-GEMM autograd, full `MegaBlocksMoeMLP` forward; both via local cpp-extension build (torch 2.10) and via the real `nix` kernel-builder artifact (`torch212-cxx11-rocm72`) loaded in the matching dev shell.


megablocks/torch-ext/megablocks/__init__.py
```diff
@@ -3,7 +3,9 @@
 
 import torch
 
-from ._ops import ops
+# Stable alias: bare `ops` is shadowed by `from . import layers` below.
+from ._ops import ops as _compiled_ops
+from . import ops
```
drbh, line 8: can you explain why we need both `_compiled_ops` and `ops` exposed?

Abdennacer-Badaoui: Inside the megablocks package there are two different "ops" things that happen to share the same name:

  1. The C++ kernels (compiled .so) : sort, histogram, etc. as raw Torch ops. Fast, no autograd, takes all the args (e.g. sort(x, end_bit, x_out, iota_out) : 4 args).
  2. A Python folder also called ops/ : wrappers around those C++ kernels that add autograd support. Same names, different signatures (e.g. sort(x, end_bit=None) : 2 args, returns a tuple).

Abdennacer-Badaoui: Before my change, the single name ops was bound to the compiled namespace, then silently overwritten to the submodule when `from . import layers` ran ; so half the file got one and half got the other depending on call order. That's the bug that broke `megablocks.sort`. Giving them distinct names makes the intent explicit and removes the order-dependence.

drbh: thanks for the explanation and the fix!

drbh (approved):
looks good to me!

PR #917, opened 2026-06-02
activation: ensure inputs are contiguous in bare activation functions

### LINKED ISSUE #106: GeLU kernel fails on non-contiguous tensors (ModernBERT)
Might be the same as https://github.com/huggingface/kernels-community/issues/76

## Summary

The `GeLU` kernel from `kernels-community/activation` produces **incorrect outputs for non-contiguous tensors**. This causes completely wrong predictions when used with ModernBERT, which uses `.chunk()` to create non-contiguous activation inputs.

## Root Cause

ModernBERT's MLP uses a gated architecture:
```python
def forward(self, hidden_states):
    input, gate = self.Wi(hidden_states).chunk(2, dim=-1)  # Creates non-contiguous tensors!
    return self.Wo(self.drop(self.act(input) * gate))
```

The `.chunk()` operation produces non-contiguous tensors. The GeLU kernel's `ops.gelu` function doesn't handle these correctly.

## Proof

```python
# Non-contiguous input (from chunk)
act_input, gate = wi_output.chunk(2, dim=-1)
print(act_input.is_contiguous())  # False
print(act_input.stride())          # (2304, 1) - not contiguous!

# Kernel fails on non-contiguous
out_kern = gelu_kernel(act_input)
out_orig = gelu_original(act_input)
print((out_orig - out_kern).abs().max())  # 17.5 ❌

# Kernel works on contiguous
out_kern = gelu_kernel(act_input.contiguous())
out_orig = gelu_original(act_input.contiguous())
print((out_orig - out_kern).abs().max())  # 0.0 ✓
```

## Environment

- Model: `answerdotai/ModernBERT-base`
- Kernel: `GeLU` from `kernels-community/activation`
- PyTorch: 2.x with CUDA

## Minimal Reproduction

```python
from transformers import AutoTokenizer, AutoModelForMaskedLM
import torch
from kernels import kernelize, Mode

model_id = "answerdotai/ModernBERT-base"
device = "cuda"

tokenizer = AutoTokenizer.from_pretrained(model_id)
text = "The capital of France is [MASK]."
inputs = tokenizer(text, return_tensors="pt")
inputs = {k: v.to(device) for k, v in inputs.items()}

# Without kernelize - CORRECT
model = AutoModelForMaskedLM.from_pretrained(model_id).to(device)
outputs = model(**inputs)
masked_index = inputs["input_ids"][0].tolist().index(tokenizer.mask_token_id)
print("Without kernelize:", tokenizer.decode(outputs.logits[0, masked_index].argmax()))
# Output: Paris ✓

# With kernelize - WRONG (due to non-contiguous tensor bug)
model2 = AutoModelForMaskedLM.from_pretrained(model_id).to(device)
model2 = kernelize(model2, mode=Mode.INFERENCE, device=device)
outputs2 = model2(**inputs)
print("With kernelize:", tokenizer.decode(outputs2.logits[0, masked_index].argmax()))
# Output: required ✗
```

## Impact

| Metric | Value |
|--------|-------|
| Max logit diff | 47.0 |
| Mean logit diff | 5.6 |
| Prediction | Completely wrong |

## Affected Models

Any model that passes non-contiguous tensors to GELUActivation, including:
- ModernBERT (uses `.chunk()` in gated MLP)
- Potentially other gated architectures

## Recommendation

- ❌ **Do not use `kernelize` on ModernBERT** until the kernel is fixed
- The `GeLU` kernel should either:
  1. Call `.contiguous()` on input before processing, or
  2. Handle strided tensors correctly in `ops.gelu`

IneshReddy249:
Fixes #106.

## Motivation

ModernBERT crashes when `kernelize()`d. Its gated MLP does `x.chunk(2, dim=-1)` and passes the non-contiguous first chunk to the activation. The bare `gelu()` in `__init__.py` doesn't call `.contiguous()` before invoking the op, so it hits the `TORCH_CHECK(input.is_contiguous())` added in #87:

```
RuntimeError: Expected input.is_contiguous() to be true, but got false.
```

The `nn.Module` wrappers in `layers.py` already guard against this. The bare functions in `__init__.py` don't — and `kernelize()` patches the bare functions, so that's the path users actually hit.

## Changes

- Add `if not x.is_contiguous(): x = x.contiguous()` to the 6 bare element-wise activation functions in `activation/torch-ext/activation/__init__.py`: `gelu`, `silu`, `gelu_tanh`, `gelu_fast`, `gelu_new`, `gelu_quick`.
- Mirrors the existing pattern from the `nn.Module` wrappers in `layers.py`.
- `*_and_mul` family is out of scope — its wrappers already have the guard.
- Pure Python; no kernel rebuild, no ABI change.

## Testing

Repro on `main` (NVIDIA A100 80GB, PyTorch 2.11.0+cu128, fp16):

```python
import torch

act = get_kernel("kernels-community/activation", version=1)
x = torch.randn(2, 128, 1536, device="cuda", dtype=torch.float16)
inp, _ = x.chunk(2, dim=-1)
assert not inp.is_contiguous()

out = torch.empty_like(inp)
act.gelu(out, inp)
# RuntimeError: Expected input.is_contiguous() to be true, but got false.
```

With the fix applied (patched bare functions overlaid onto the Hub kernel for verification), the same ModernBERT chunk pattern produces matching outputs across all six activations vs the `torch.nn.functional` reference, and the contiguous path is unchanged:

```
gelu         max diff 0.0
silu         max diff 0.0
gelu_tanh    max diff 0.0
gelu_fast    max diff 0.0
gelu_new     max diff 0.0
gelu_quick   max diff 0.0
gelu contig  max diff 0.0   (no regression)
```

### DIFF
--- a/activation/torch-ext/activation/__init__.py
+++ b/activation/torch-ext/activation/__init__.py
@@ -31,30 +31,43 @@ def fatrelu_and_mul(out: torch.Tensor, x: torch.Tensor, threshold: float = 0.0) 
 
 
 def gelu(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
+    if not x.is_contiguous():
+        x = x.contiguous()
     ops.gelu(out, x)
     return out
 
+
 def silu(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
+    if not x.is_contiguous():
+        x = x.contiguous()
     ops.silu(out, x)
     return out
 
 
 def gelu_tanh(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
+    if not x.is_contiguous():
+        x = x.contiguous()
     ops.gelu_tanh(out, x)
     return out
 
 
 def gelu_fast(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
+    if not x.is_contiguous():
+        x = x.contiguous()
     ops.gelu_fast(out, x)
     return out
 
 
 def gelu_new(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
+    if not x.is_contiguous():
+        x = x.contiguous()
     ops.gelu_new(out, x)
     return out
 
 
 def gelu_quick(out: torch.Tensor, x: torch.Tensor) -> torch.Tensor:
+    if not x.is_contiguous():
+        x = x.contiguous()
     ops.gelu_quick(out, x)
     return out
 


IneshReddy249:
@danieldk apologies, missed the vouching policy when opening this. CONTRIBUTING.md doesn't mention it, so I only saw the requirement after the bot closed the PR.

The diff is 13 lines of `.contiguous()` guards in `__init__.py`, mirroring the pattern that's already in `layers.py`. Tested on A100, repro and verification in the PR body. If there's a path to getting vouched and reviewed, happy to follow it. Otherwise feel free to land the diff under your own commit, no attribution needed.

PR #919, opened 2026-06-03
package ops inside flash attention separately.

### LINKED ISSUE #900: Package flash-attn Triton ops (cross_entropy, rotary, etc.) as a kernel
While integrating kernels for [dexmal/dexbotic#94](https://github.com/dexmal/dexbotic/issues/94), we noticed that the project imports flash_attn.losses.cross_entropy ([source](https://github.com/dexmal/dexbotic/blob/3f287c9d05520f2a19d9e3041789395496e9f423/dexbotic/exp/rl/rl_trainer.py#L16)), which relies on kernels bundled inside the flash-attn package that we don't currently expose.

## Proposal

Package the Triton ops from flash-attention as a standalone kernel (e.g. flash-attn-triton):

- Minimum: [flash_attn/ops/triton](https://github.com/Dao-AILab/flash-attention/tree/main/flash_attn/ops/triton) (includes cross-entropy, layer norm, rotary, etc.)
- Ideal: The full [flash_attn/ops](https://github.com/Dao-AILab/flash-attention/tree/main/flash_attn/ops) folder, covering all the utility kernels people commonly import from flash-attn.

## Motivation

These ops are used fairly widely across training codebases. Packaging them independently would let projects depend on just the kernels they need without pulling in the full flash-attn build (which is often a blocker for integration).

## Notes

- Some of these may already overlap with existing kernel repos (e.g. the RoPE kernel may be based on the flash-attn rotary op) — worth auditing for deduplication.

sayakpaul:
Package the ops (rotary, cross-entropy loss, etc.) from Flash Attention. 

Doesn't include `linear.py` as it needs `triton.ops.matmul_perf_model`. This was removed in Triton ≥ 3.0. Didn't include `mlp.py`, it needs the compiled `fused_dense_lib`. Not sure how to include them.

Fixes https://github.com/huggingface/kernels-community/issues/900


flash-attn-ops-plan.md
```diff
@@ -0,0 +1,370 @@
+# Plan: `flash-attn-ops` — a `kernels`-compliant package for flash-attn's Triton ops
```
drbh: was this intended on being included in the PR? this looks like a dev artifact we may want to remove

sayakpaul: Removed it. Just used it to get the lay of the land on what's needed on my end.

rycerzes (commented):
Hi @sayakpaul, thank you for adding the updated flash attn Triton ops. Shouldn't the old [rotary](https://github.com/huggingface/kernels-community/tree/main/rotary) be replaced by the new rotary kernel since the old one didn't support partial RoPE, and would result in corruption and over rotation

sayakpaul:
> Shouldn't the old [rotary](https://github.com/huggingface/kernels-community/tree/main/rotary) be replaced by the new rotary kernel since the old one didn't support partial RoPE, and would result in corruption and over rotation

Can you point me to the new ones?

rycerzes:
> Can you point me to the new ones? 

https://github.com/Dao-AILab/flash-attention/blob/d80a77103021c4e980f8cbbf85774f6a19e6474a/flash_attn/ops/triton/rotary.py#L13

It's already in this PR. In the flash-attn repo, they removed the `csrc/rotary` at commit [f28841d](https://github.com/Dao-AILab/flash-attention/commit/f28841db5043c6a329869b6c3e4e3f5f5ebdc1a0)

sayakpaul:
Maybe we should do that in a follow-up? Open an issue thread for that?

drbh (approved):

sayakpaul:
Failing tests are unrelated.

sayakpaul:
@drbh any reason why the bot didn't act on the comment above?

drbh:
yea I think its because the comment above "` /kernel-bot merge-and-upload flash-attn-ops`" contains a space before the `/kernel-bot` command and the script only matches if the text starts with `/kernel-bot`. we may want to trim whitespace to avoid this issue in the future

PR #922, opened 2026-06-04
feat: allow kernel-bot to run security analysis.

### LINKED ISSUE #901: Security audit workflow: support external PRs without forking
Description:

Currently the security audit workflow fails on PRs from forks because:
1. The workflow cannot safely run automatically on external PRs (risk of adversarial prompt injection via the diff).
2. The dispatch-based trigger also fails for similar reasons.

Current workaround: A repo maintainer must fork the PR (check it out and recreate it from a non-fork branch) so that actions run.

Desired end state:

- Do not run any job by default on external PRs.
- A repo maintainer does a first sanity check of the PR, then invokes kernel-bot to do the review (e.g. /kernel-bot security-review).
- This avoids the convoluted fork-and-recreate workflow while maintaining security (a human validates no jailbreak instructions before the bot processes the diff).

Blocked on: Confirming the security model of kernel-bot for third-party submissions (waiting on @<maintainer>).

References:

- https://github.com/huggingface/kernels-community/actions/runs/26636371117/job/78499706172?pr=889
- https://github.com/huggingface/kernels-community/pull/680
- https://github.com/huggingface/kernels/blob/53b30e0df1423f3483980a11c5b0e0d86bffada3/.github/workflows/security-audit.yml#L33

Needs to be propagated to https://github.com/huggingface/kernels/ too.

sayakpaul:
Fixes https://github.com/huggingface/kernels-community/issues/901

Additionally,

* Modifies the security audit analysis workflow to _only_ trigger on "workflow dispatch". So, the idea is that a maintainer would do `kernel-bot security` after initially vetting that a PR doesn't have any jailbreaking instructions.
* Removes the code to post to Slack. Instead, the action comments on a concerned PR. That way, things remain transparent and self-contained.


sayakpaul:
I am afraid we cannot test it without merging to `main` similar to how these kinds of things were tested in the past.

.github/scripts/pr_comment_kernel_bot.py
```diff
@@ -22,6 +22,7 @@
     "merge-and-upload": {"admin"},
     "release": {"admin"},
 }
+SECURITY_WORKFLOW = "security-audit.yml"
```
drbh: I think this should be in `dispatch.py` since that is where we list the other workflows we trigger

.github/scripts/pr_comment_kernel_bot.py
```diff
@@ -431,6 +451,39 @@ def comment_has_only_supported_characters(comment: str):
...
+        token,
+        SECURITY_WORKFLOW,
+        default_branch,
+        {
+            "pr_number": str(issue_number),
+            "dispatch_key": dispatch_key,
+            "head_sha": pr_head_sha or "",
+        },
+    )
+    return DispatchResult(kernel_name="security", dispatch_key=dispatch_key)
```
drbh: this logic should probably also be in the dispatch script too, so we follow the existing patterns

drbh (changes requested):
looks good overall, but I think we should move the dispatch logic into the `dispatch.py` file and update the `do_dispatch_release` function to take any new args needed to conditionally trigger the security workflow. 

**also I think maybe some of the existing function names could be improved (ie maybe dispatch_release should be dispatch?) if you want to make any changes to that feel free to update. however we can totally improve in another PR, and was just a thought since you're touching these files

sayakpaul:
@drbh your comments are addressed now. PTAL.

drbh (approved):

PR #925, opened 2026-06-04
ci: merge-and-upload branch default should be unset

danieldk:
Otherwise we only upload to the default branch (`main`) and the version (branch) in `build.toml` is not respected.


sayakpaul (approved):
LGTM. I think it could be nice to make the branch a required parameter when uploading to prevent cases like this.

PR #929, opened 2026-06-04
rotary: Refactor split into separate files + warning in readme

vasqu:
As per title, first attempt at splitting this into proper files. Also added a warning to the readme that this is from the FA repo but no longer maintained, let's see when we move away from that implementation


rotary/torch-ext/rotary/layers.py
vasqu, line 1: This is the split

rotary/CARD.md
```diff
@@ -3,6 +3,9 @@ library_name: kernels
 {% if license %}license: {{ license }}
 {% endif %}---
 
+> [!WARNING]
```
vasqu, line 6: Small warning to make sure users are aware

PR #932, opened 2026-06-05
[security] use env variables through env in the workflows

sayakpaul:


danieldk (dismissed):
Looks, good. Good to merge if CI runs successfully.

PR #938, opened 2026-06-08
finegrained-fp8: fix build 

IlyasMoutawwakil:
## Related issue

Thought the build was successful but it wasn't (didn't know we had to check the action) 😅

Closes #

## Changes

-

## Testing

-

## Checklist

- LLM disclosure:

### DIFF
--- a/finegrained-fp8/torch-ext/finegrained_fp8/utils.py
+++ b/finegrained-fp8/torch-ext/finegrained_fp8/utils.py
@@ -81,8 +81,18 @@ def grouped_tile_layout(
 
 @functools.cache
 def get_active_device_type() -> str:
-    """Active torch device type for the current Triton backend (``"cuda"``, ``"xpu"``, ...)."""
-    return triton.runtime.driver.active.get_active_torch_device().type
+    """Active torch device type for the current Triton backend (``"cuda"``, ``"xpu"``, ...).
+
+    Falls back to ``"cuda"`` when no driver is loaded — Triton raises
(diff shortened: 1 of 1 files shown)


finegrained-fp8/torch-ext/finegrained_fp8/utils.py
```diff
@@ -81,8 +81,18 @@ def grouped_tile_layout(
...
+    Falls back to ``"cuda"`` when no driver is loaded — Triton raises
+    ``RuntimeError: 0 active drivers ([])`` on driverless build boxes, and the
+    autotune-config builder is evaluated at module-import time under the
+    ``@triton.autotune`` decorator (no kernel launches there, so the default is
+    only used to shape the config list).
+    """
+    try:
+        return triton.runtime.driver.active.get_active_torch_device().type
+    except RuntimeError:
+        return "cuda"
```
vasqu, line 95: Don't really understand why we fallback and ignore the runtime error tbh 😅 shouldnt we just let it error out?

vasqu (approved):
Just a small question but build looks successful 🫡

finegrained-fp8/torch-ext/finegrained_fp8/utils.py
IlyasMoutawwakil: this is because the autotune decorator is also executed during "build" where we don't have a gpu. that's why previous build failed.

vasqu: Ah makes sense (i shouldve read the docstring 😬)

PR #943, opened 2026-06-10
rotary: Add rotary for CPU backend

jiqing-feng:
## Summary

This PR adds a CPU backend for the `rotary` kernel so that
`apply_rotary` / `apply_rotary_transformers` run natively on CPU (previously only
`cuda` and `xpu` backends existed, and CPU tensors fell back to the eager
`apply_rotary_pos_emb` in transformers).

The implementation follows the same fp32-accumulation semantics as the existing
CUDA/XPU kernels:

```
conj == false:  out1 = x1*cos - x2*sin    out2 = x1*sin + x2*cos
conj == true :  out1 = x1*cos + x2*sin    out2 = x2*cos - x1*sin
```

All arithmetic is done in fp32 and rounded back to the input dtype, so the kernel
is actually *more* accurate than the eager bf16 path (which accumulates in bf16).

## What's added

New `rotary-cpu/` sources plus `build.toml` / `torch_binding.cpp` wiring:

- `cpu_features.hpp` — runtime AVX512 detection (CPUID + XCR0), in its own
  `rotary_cpu` namespace.
- `rotary_cpu.cpp` — dispatcher. Picks the AVX512 path for contiguous-last-dim
  BF16 4-D tensors with `cos`/`sin` broadcast over heads; otherwise a generic
  ATen fp32 fallback that works for any dtype/layout on any CPU.
- `rotary_avx512.cpp` — vectorized BF16 implementation using `at::vec`.
- `rotary_cpu_torch.cpp` — global `_apply_rotary` entry point used by the binding.
- `build.toml` — adds `"cpu"` to `backends` and two CPU `[kernel.*]` sections
  (base + AVX512 with its own `cxx-flags`).
- `torch_binding.cpp` — allows `kCPU` in `CHECK_DEVICE` and registers the
  `torch::kCPU` impl.

The Python wrapper is unchanged: PyTorch dispatches CPU tensors to the new impl
automatically.

## Correctness

Validated against the eager transformers rotary (`rotate_half`) over bf16 / fp16
/ fp32, multiple `head_dim`s (64/96/128), the `conj=true` path, and both
contiguous and transposed (real-attention-layout) `q`/`k`. Using an fp32 ground
truth, the kernel's error is consistently *below* the eager-bf16 error, confirming
the fp32 accumulation is a small accuracy win on top of the speedup.

## Performance

Element-wise / memory-bound op. Measured against the eager baseline on a single
NUMA node (Intel Xeon 6, bf16, 32 threads, `torch.utils.benchmark` median),
Qwen2.5-0.5B rotary shapes (14 q-heads / 2 kv-heads / head_dim 64):

| case               | speedup |
|--------------------|---------|
| decode bs1         | ~1.8x   |
| decode bs16        | ~2.3x   |
| decode bs64        | ~1.9x   |
| prefill bs1 s512   | ~2.3x   |
| prefill bs1 s2048  | ~5.7x   |
| prefill bs8 s512   | ~3.9x   |

End-to-end (Qwen2.5-0.5B) the change is small but consistently positive
(~4–6% decode, ~10% large-batch prefill), as expected: rotary is a tiny fraction
of per-token cost, which is dominated by matmuls.

## How the `cpu-kernels` skill was used

I followed the skill's two-phase workflow:

- **Phase 1 (correctness):** generic ATen fp32 fallback first, then the AVX512
  tier in its own translation unit with its own `cxx-flags`, using the required
  `cpu_features.hpp` runtime-dispatch pattern and `at::vec::Vectorized` with
  `-DCPU_CAPABILITY_AVX512` so the BF16 conversions actually lower to AVX512
  (rather than silently falling back to the scalar `vec_base.h` path).
- **Phase 2 (performance):** starting from a correct AVX512 kernel, I profiled
  via the roofline lens the skill recommends (this op is memory-bound) and then,
  following the decision tree's "prefer a structural change over knob-sweeping"
  guidance, restructured the parallel loop from per-`(batch, head, seq)` rows to
  per-`(batch, seq)` positions with `cos`/`sin` converted once and reused across
  all heads. Because `cos`/`sin` broadcast over heads, this removes redundant
  loads + bf16→fp32 conversions and was the main win on the bandwidth-bound
  prefill cases (e.g. `bs1 s2048` 3.8x → 5.7x). A follow-up FMA trial was neutral
  (the compiler already contracted to FMA) and kept only for robustness.

> Test harnesses live outside this repo; only kernel sources, `build.toml`, and
> `torch_binding.cpp` are changed here.

### DIFF
--- a/rotary/build.toml
+++ b/rotary/build.toml
@@ -5,6 +5,7 @@ version = 1
 backends = [
     "cuda",
     "xpu",
+    "cpu",
 ]
 
 
@@ -26,3 +27,35 @@ src = [
 backend = "cuda"
 depends = ["torch"]
 src = ["rotary/rotary_cuda.cu"]
+
+[kernel.rotary_cpu]
+backend = "cpu"
+depends = ["torch"]
+include = ["rotary-cpu"]
+src = [
+    "rotary-cpu/rotary_cpu_torch.cpp",
+    "rotary-cpu/rotary_cpu.cpp",
+    "rotary-cpu/rotary_cpu.hpp",
+    "rotary-cpu/cpu_features.hpp",
+]
+
+[kernel.rotary_cpu_avx512]
+backend = "cpu"
+depends = ["torch"]
+include = ["rotary-cpu"]
+cxx-flags = [
+    "-O3",
+    "-fopenmp",
+    "-mf16c",
+    "-mavx512f",
+    "-mavx512bf16",
+    "-mavx512vl",
+    "-mavx512dq",
+    "-mavx512bw",
+    "-DCPU_CAPABILITY_AVX512",
+    "-DCPU_CAPABILITY=AVX512",
+]
+src = [
+    "rotary-cpu/rotary_avx512.cpp",
+    "rotary-cpu/rotary_avx512.hpp",
+]
--- a/rotary/flake.nix
+++ b/rotary/flake.nix
@@ -8,5 +8,15 @@
     kernel-builder.lib.genKernelFlakeOutputs {
       inherit self;
       path = ./.;
+
+    torchVersions =
+      let
+        # For CPU builds, only x86_64-linux is currently supported.
+        cpuSupported = version: system: !(version ? "cpu") || system == "x86_64-linux";
+      in
+      allVersions:
+      builtins.map (
+        version: version // { systems = builtins.filter (cpuSupported version) version.systems; }
+      ) allVersions;
     };
 }
--- a/rotary/rotary-cpu/cpu_features.hpp
+++ b/rotary/rotary-cpu/cpu_features.hpp
@@ -1,0 +1,59 @@
+#pragma once
+
+#ifdef _MSC_VER
+#include <intrin.h>
+#else
+#include <cpuid.h>
+#endif
+
+namespace rotary_cpu {
+
+// Runtime CPU feature detection. Used to pick the AVX512 path only on hardware
+// that actually supports it; otherwise the generic ATen fallback is used.
+class CPUFeatures {
+public:
+  static bool hasAVX512() {
+    static bool supported = checkAVX512();
+    return supported;
+  }
+
+private:
+  static bool checkAVX512() {
+#ifdef _MSC_VER
+    int cpu_info[4];
+    __cpuid(cpu_info, 0);
+    if (cpu_info[0] < 7)
+      return false;
+    __cpuidex(cpu_info, 7, 0);
+    bool avx512f = (cpu_info[1] & (1 << 16)) != 0; // EBX bit 16
+    if (!avx512f)
+      return false;
+    __cpuid(cpu_info, 1);
+    bool osxsave = (cpu_info[2] & (1 << 27)) != 0; // ECX bit 27
+    if (!osxsave)
+      return false;
+    unsigned long long xcr0 = _xgetbv(0);
+    return ((xcr0 & 0xE6ULL) == 0xE6ULL);
+#else
+    unsigned int eax, ebx, ecx, edx;
+    if (__get_cpuid_max(0, nullptr) < 7)
+      return false;
+    __cpuid_count(7, 0, eax, ebx, ecx, edx);
+    bool avx512f = (ebx & (1 << 16)) != 0; // EBX bit 16
+    if (!avx512f)
+      return false;
+    if (__get_cpuid(1, &eax, &ebx, &ecx, &edx) == 0)
+      return false;
+    bool osxsave = (ecx & (1 << 27)) != 0; // ECX bit 27
+    if (!osxsave)
+      return false;
+    unsigned int xcr0_lo = 0, xcr0_hi = 0;
+    __asm__ volatile("xgetbv" : "=a"(xcr0_lo), "=d"(xcr0_hi) : "c"(0));
+    unsigned long long xcr0 =
+        ((unsigned long long)xcr0_hi << 32) | xcr0_lo;
+    return ((xcr0 & 0xE6ULL) == 0xE6ULL);
+#endif
+  }
+};
+
+} // namespace rotary_cpu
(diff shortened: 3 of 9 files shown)


jiqing-feng:
Benchmark script:
```python
from pathlib import Path

import torch
import torch.utils.benchmark as benchmark
from kernels import get_local_kernel

KERNEL_DIR = Path("/home/jiqing/dqhyr0kf3dgv4nr16fj5pyzwb47gj2aq-rotary-torch-ext")

rotary = get_local_kernel(KERNEL_DIR, backend="cpu")
print(f"loaded kernel: {rotary}  threads={torch.get_num_threads()}\n")


# ---- torch eager baseline (transformers apply_rotary_pos_emb) ----
def rotate_half(x):
    x1 = x[..., : x.shape[-1] // 2]
    x2 = x[..., x.shape[-1] // 2 :]
    return torch.cat((-x2, x1), dim=-1)


def eager_rope(q, k, cos, sin, unsqueeze_dim=1):
    cos = cos.unsqueeze(unsqueeze_dim)
    sin = sin.unsqueeze(unsqueeze_dim)
    return (q * cos) + (rotate_half(q) * sin), (k * cos) + (rotate_half(k) * sin)


def make_inputs(bs, qh, kvh, seq, head_dim, dtype=torch.bfloat16):
    # q/k laid out like real attention: (bs, seq, heads, hd) -> transpose
    q = torch.randn(bs, seq, qh, head_dim, dtype=dtype).transpose(1, 2)
    k = torch.randn(bs, seq, kvh, head_dim, dtype=dtype).transpose(1, 2)
    pos = torch.arange(seq, dtype=torch.float32)
    inv = 1.0 / (10000 ** (torch.arange(0, head_dim, 2).float() / head_dim))
    emb = torch.cat((torch.outer(pos, inv),) * 2, dim=-1)
    cos = emb.cos().to(dtype).unsqueeze(0).expand(bs, seq, head_dim).contiguous()
    sin = emb.sin().to(dtype).unsqueeze(0).expand(bs, seq, head_dim).contiguous()
    return q, k, cos, sin


def med_us(fn, *args):
    t = benchmark.Timer(
        stmt="fn(*args)",
        globals={"fn": fn, "args": args},
        num_threads=torch.get_num_threads(),
    )
    return t.blocked_autorange(min_run_time=2.0).median * 1e6


# Qwen2.5-0.5B: 14 q heads, 2 kv heads, head_dim 64
CASES = [
    ("decode bs1", 1, 14, 2, 1, 64),
    ("decode bs4", 4, 14, 2, 1, 64),
    ("decode bs16", 16, 14, 2, 1, 64),
    ("decode bs64", 64, 14, 2, 1, 64),
    ("prefill bs1 s128", 1, 14, 2, 128, 64),
    ("prefill bs1 s512", 1, 14, 2, 512, 64),
    ("prefill bs1 s2048", 1, 14, 2, 2048, 64),
    ("prefill bs8 s128", 8, 14, 2, 128, 64),
    ("prefill bs8 s512", 8, 14, 2, 512, 64),
]

print(f"{'case':<20}{'eager(us)':>12}{'kernel(us)':>12}{'speedup':>10}")
print("-" * 54)
for name, bs, qh, kvh, seq, hd in CASES:
    q, k, cos, sin = make_inputs(bs, qh, kvh, seq, hd)
    # correctness sanity vs fp32 ground truth
    gt_q, gt_k = eager_rope(q.float(), k.float(), cos.float(), sin.float())
    kq, kk = rotary.apply_rotary_transformers(q, k, cos, sin)
    err = max(
        (kq.float() - gt_q).abs().max().item(),
        (kk.float() - gt_k).abs().max().item(),
    )
    assert err < 5e-2, f"{name}: correctness err={err:.2e}"
    e = med_us(eager_rope, q, k, cos, sin)
    m = med_us(rotary.apply_rotary_transformers, q, k, cos, sin)
    print(f"{name:<20}{e:>12.2f}{m:>12.2f}{e / m:>9.2f}x")
```

jiqing-feng:
Follow-up optimization opened separately as https://github.com/jiqing-feng/kernels-community/pull/1, so it can be reviewed/merged independently of #943.

**What it changes**

One-function change in the shared `apply_rotary_transformers` wrapper (`torch-ext/rotary/__init__.py`): instead of cloning `q`/`k` and rotating in place, it writes into fresh `torch.empty_like` buffers and reads `q`/`k` directly, relying on the kernel's `out != in` support. This removes the extra full read+write of cloning before an in-place rotation, ~halving the wrapper's memory traffic on this bandwidth-bound op.

**Blast radius**

- It's the **shared** wrapper, so it touches `cpu`, `cuda` and `xpu`.
- It now calls the kernels with `out != in` (previously always in-place). This is safe **by construction**: both `rotary_cuda.cu` and the XPU kernel are `TensorIterator` multi-output elementwise kernels whose lambda takes inputs **by value**, so `out != in` and `out == in` are handled identically; the CPU kernel also reads inputs into registers before storing. No kernel assumes in-place.
- Output is **bit-identical** to the old clone path, so it's behavior-preserving with no kernel changes.

**Validation (CPU, single-NUMA Xeon 6, bf16)**

- Correctness: kernel output `max|diff| == 0` vs the clone path; greedy-128 `new == old` is `True` on Qwen2.5-3B.
- Op-level vs the old clone wrapper: decode bs16 **2.0×**, bs64 **2.6×**, prefill bs1 s512 **2.1×**, bs8 s512 **1.3×**. Caveat: tiny `decode bs1`/`bs4` regress slightly (~0.8×) since the noclone path does a few more Python view ops, which dominates at that size; crossover is around bs16.
- E2E (Qwen2.5-3B, three-way same-process): `new / old` is within noise on every case (0.99–1.03×) — **no e2e regression**; both stay a small net win over eager.

**Question for maintainers**

Since this is a shared-wrapper change, would you prefer to (1) merge it as the separate follow-up PR after a quick CUDA smoke test on your side, (2) fold it into #943, or (3) hold it until the CPU backend lands? I don't have CUDA hardware to smoke-test myself, but per the `TensorIterator` by-value semantics above I believe it's safe on both.


Hi @danieldk . Would you please review it? Thanks!

danieldk (approved):

PR #948, opened 2026-06-11
liger-kernels: sync changes for transformers

vasqu:
As per title, I need a few changes for transformers to work properly
- Signature fixes
- Init fixes
- Torch compile fixes

NOTE: This discards a few non-used modules/funcs in layers. We will rebuild them gradually if needed. But this is breaking either way so we will need a new version regardless


liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -1,360 +1,142 @@
...
-from .qwen2vl_mrope import LigerQwen2VLMRopeFunction
 from .rms_norm import LigerRMSNormFunction
 from .rope import LigerRopeFunction
 from .swiglu import LigerSiLUMulFunction
 from .tiled_mlp import apply_tiled_mlp
-from .tvd import LigerTVDLossFunction
 
 
+# NOTE: Not compile-friendly --> large deviations to the original implementation under compile
 class LigerRMSNorm(nn.Module):
```
vasqu, line 17: There are a few rms norm alts, will check if this is worth it to add, e.g. for gemma iirc

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -1,360 +1,142 @@
...
-            self.num_groups,
-            self.variance_epsilon,
-        )
+    def extra_repr(self):
+        return f"{tuple(self.weight.shape)}, eps={self.variance_epsilon}"
 
-    def extra_repr(self) -> str:
-        return f"num_channels={self.num_channels}, num_groups={self.num_groups}, eps={self.variance_epsilon}"
 
+class LigerLinear(nn.Module):
```
vasqu, line 36: This can be used to exchange a torch native nn.Linear but using chunked/tiled forwards and backwards. This can be seen as something similar to what the other tiled MLPs do. This won't land on main transformers immediately but I'm optimistic that it should and can

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -1,360 +1,142 @@
...
 
 
-class LigerTiledSwiGLUMLP(nn.Module):
+class LigerTiledGEGLUMLP(nn.Module):
     gate_proj: nn.Linear
     up_proj: nn.Linear
     down_proj: nn.Linear
-    num_shards: int
 
-    def _mlp_forward(self, module, x):
```
vasqu, line 340: NOTE: this adds an "attribute" / self reference to something the original modules do not have --> cannot be properly exchanged via kernelize as it fails validation

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -1,360 +1,142 @@
...
+            fn=_mlp_forward,
             mlp_module=self,
             x=x,
-            num_shards=self.num_shards,
+            num_shards=None,
             compute_params=compute_params,
         )
 
 
+def liger_rotary_pos_emb(
```
vasqu, line 129: Could be used instead of rotary tbh, would need to check between FA ops and this what's more worthwhile but it was simple enough to add so why not

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -407,23 +189,33 @@ def liger_fused_linear_cross_entropy(
     )
 
 
+# NOTE: We have this intentional graph break as we encounter issues such as IMAs and Cublas errors.
+#       We know that this is an optimized kernel already so there is less ways to
+#       fuse it either way; we rely on torch compile to go through the base model to optimize.
+@torch.compiler.disable
```
vasqu, line 195: Let's keep this for now, adding fake ops etc will take a bit more effort. We can always lift this restriction later tbh

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -435,80 +227,38 @@ def LigerForCausalLMLoss(
     shift_labels = shift_labels.view(-1).to(hidden_states.device)
 
     reduction = "sum" if num_items_in_batch is not None else "mean"
-    result = liger_fused_linear_cross_entropy(
+    loss = liger_fused_linear_cross_entropy(
         hidden_states,
         lm_head_weight,
         shift_labels,
+        bias=lm_head_bias,
```
vasqu, line 234: Was missed --> a few lm heads do have a bias it seems

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -435,80 +227,38 @@ def LigerForCausalLMLoss(
...
 __all__ = [
     "LigerRMSNorm",
-    "LigerLayerNorm",
-    "LigerGroupNorm",
-    "LigerDyT",
-    "LigerCrossEntropyLoss",
-    "LigerFusedLinearCrossEntropyLoss",
-    "LigerJSD",
-    "LigerKLDIVLoss",
-    "LigerTVDLoss",
```
vasqu, line 504: Like mentioned we can gradually readd some of these when actually needed. Let's keep it simple for now

liger-kernels/torch-ext/liger_kernels/tiled_mlp.py
```diff
@@ -62,21 +62,27 @@ def backward(ctx, *grads) -> tuple:
...
         x_requires_grad = x.requires_grad
         x = x.detach()
         # detach() unsets x.requires_grad, so restore it
         x.requires_grad_(x_requires_grad)
 
-        # x.shape could be [bs, seqlen, hidden_size] or [seqlen, hidden_size] (moe experts)
-        hidden_size = x.shape[-1]
+        # x.shape could be [bs, seqlen, in_features] or [seqlen, in_features] (moe experts)
+        in_features = x.shape[-1]
+        out_features = grad_output.shape[-1]
```
vasqu, line 74: This is to support any mlp going `in->out`; the basis here was that we have `in->in` shapes which does not hold for anything arbitrary

vasqu (commented):
Self-review to clarify a few things

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -435,80 +227,38 @@ def LigerForCausalLMLoss(
...
-    sin: torch.Tensor,
-    mrope_section,
-    unsqueeze_dim: int = 1,
-) -> Tuple[torch.Tensor, torch.Tensor]:
-    """Apply Qwen2-VL multimodal rotary positional embedding (M-RoPE) to ``q`` and ``k``."""
-    return LigerQwen2VLMRopeFunction.apply(q, k, cos, sin, mrope_section, unsqueeze_dim)
+# Add torch compile support for functions - in this case it's to allow this to be used
+# but the function itself will not be compiled (see the note at the function)
+liger_rotary_pos_emb.can_torch_compile = True
+LigerForCausalLMLoss.can_torch_compile = True
```
danieldk, line 252: Maybe it's worth writing these as layers now already, since we are going to deprecate `FuncRepository`?

vasqu: Tried something along
```py
class LigerForCausalLMLoss(nn.Module):
    # NOTE: We have this intentional graph break as we encounter issues such as IMAs and Cublas errors.
    #       We know that this is an optimized kernel already so there is less ways to
    #       fuse it either way; we rely on torch compile to go through the base model to optimize.
    @torch.compiler.disable
    def forward(
        self,
        logits: None,  # to match transformers signature
        labels: torch.Tensor,
        vocab_size: int,  # to match transformers signature
        num_items_in_batch: Optional[int] = None,
        ignore_index: int = -100,
        shift_labels: Optional[torch.Tensor] = None,
        hidden_states: torch.Tensor | None = None,
        lm_head_weight: torch.Tensor | None = None,
        lm_head_bias: torch.Tensor | None = None,
        **kwargs,
    ):
        # To match signature we hide these behind the kwargs but we expect a few kwargs to exist
        hidden_size = kwargs.pop("hidden_size", None)
        final_logit_softcapping = kwargs.pop("final_logit_softcapping", None)

        if hidden_size is None or hidden_states is None or lm_head_weight is None:
            raise ValueError(
                f"`LigerForCausalLMLoss` requires the LLM's weight (found `{lm_head_weight is not None}`),"
                f"the last hidden state (found `{hidden_states is not None}`), and the `hidden_size`"
                f"(found `{hidden_size is not None}`). Please make sure to pass the necessary kwargs."
            )

        applicable_params = inspect.signature(liger_fused_linear_cross_entropy).parameters
        kwargs = {k: v for k, v in kwargs.items() if k in applicable_params}

        if shift_labels is None:
            labels = nn.functional.pad(labels, (0, 1), value=ignore_index)
            shift_labels = labels[..., 1:].contiguous()

        hidden_states = hidden_states.view(-1, hidden_size)
        shift_labels = shift_labels.view(-1).to(hidden_states.device)

        reduction = "sum" if num_items_in_batch is not None else "mean"
        loss = liger_fused_linear_cross_entropy(
            hidden_states,
            lm_head_weight,
            shift_labels,
            bias=lm_head_bias,
            reduction=reduction,
            ignore_index=ignore_index,
            softcap=final_logit_softcapping,
            return_token_accuracy=False,
            return_predicted_tokens=False,
            **kwargs,
        )

        if reduction == "sum":
            loss = loss / num_items_in_batch

        return loss
```
but the signatures don't match because we do simple `func_sig = inspect.signature(func)` which does not work atm. So we will need https://github.com/huggingface/kernels/pull/622 to land first (as it properly checks whether something is a class or fn). 

We can update afterwards, wdyt?

danieldk: Sounds good.

danieldk (approved):

PR #949, opened 2026-06-11
rotary: Refactor split into separate files + warning in readme

vasqu:
As per title, successor to #929 but using direct branches and not a fork to have proper CI

- Adds a warning to the card as the kernel is outdated / not maintained anymore
- Split the func specifics to `layer` (should be done sooner or later)


danieldk:
The build currently fails:

```
       >   File "/nix/store/009nfcf16mly93zr54yywrjisv1ngpbj-rotary-torch-ext/torch212-cxx11-cu126-x86_64-linux/__init__.py", line 4, in <module>
       >     from .layers import apply_rotary_transformers
       >   File "/nix/store/009nfcf16mly93zr54yywrjisv1ngpbj-rotary-torch-ext/torch212-cxx11-cu126-x86_64-linux/layers.py", line 3, in <module>
       >     from . import apply_rotary
       > ImportError: cannot import name 'apply_rotary' from '_rotary_cuda_73f4e9f' (/nix/store/009nfcf16mly93zr54yywrjisv1ngpbj-rotary-torch-ext/torch212-cxx11-cu126-x86_64-linux/__init__.py)
 ```

**Comment by vasqu** [comment]
Will check, probably some wrong relative import. Should be straightforward

File: rotary/torch-ext/rotary/__init__.py [hunk 1]
```diff
@@ -16,41 +16,5 @@ def apply_rotary(
...
-    apply_rotary(q1, q2, cos, sin, q1, q2, False)
-    apply_rotary(k1, k2, cos, sin, k1, k2, False)
-    return q_rotated, k_rotated
-
-
-# Add torch compile support for functions
-apply_rotary_transformers.can_torch_compile = True
-
-
+# Keeping `apply_rotary` for BC
```
vasqu: Tbh, not sure whether we really should have kept the original one as it's not really used / or should not be used? Maybe we should bump the version instead and leave it out?

drbh (dismissed):
changes lgtm, just kicked off a build to make sure there are not more build issues

drbh (approved):

PR #953, opened 2026-06-15
aiter-kernels: add bundled AITER Triton kernels for AMD ROCm

Abdennacer-Badaoui:
Adds a new `aiter-kernels/` package that bundles the Triton subset of upstream `ROCm/aiter` into a single Hugging Face Hub kernel, mirroring the same approach used for liger-kernels. The package vendors `aiter/ops/triton/**` (minus Flash Attention, which continues to live in `kernels-community/aiter-flash-attn`) under `aiter_kernels.*` with a small `_aiter_compat/` shim for the few cross-tree dependencies (`dtypes, chip_info, torch_compile_guard, triton_metadata_redirect`). build.toml uses` pyext = ["py", "pyi", "json"]` so the 300+ per-arch autotuner JSON configs ship to the Hub, and the flake pins the` huggingface/kernels/torch-noarch-pyext` kernel-builder branch. An `AGENTS.md` entry codifies the sync workflow for future releases. We're starting with just the Triton kernels — HIP/CK kernels are not in scope yet.


aiter-kernels/build.toml
```diff
@@ -0,0 +1,13 @@
+[general]
+name = "aiter-kernels"
+version = 1
+license = "MIT"
+backends = ["rocm"]
+
+[general.hub]
+repo-id = "kernels-community/aiter-kernels"
```
sayakpaul, line 8: Will these kernels be upstreamed to `transformers`? If not, would it make sense to maintain these in a separate repo, set up CI, etc.?

Additionally, do we need to define any specific capabilities for rocm or are these of "no-arch" flavor?

Abdennacer-Badaoui: Some of the kernels can be used in Transformers, although there aren't many relevant ones at the moment.

The long-term goal is to have the repository hosted under the AMD organization namespace. For now, I can create it under my namespace if that works for you. Once we get a response from the AMD team, we can transfer it to their namespace.

Abdennacer-Badaoui: > Additionally, do we need to define any specific capabilities for rocm or are these of "no-arch" flavor?

No, they are of no-arch flavor.

sayakpaul (approved):

PR #954, opened 2026-06-15
finegrained-fp8: mxfp8-optimized kernels recipes

IlyasMoutawwakil:
## Important: Read before submitting

> **New contributors:** Please open or comment on an issue **before** submitting a
> PR to discuss the change you'd like to make. This helps us align on approach and
> avoids wasted effort on changes we may not be able to merge. Please only create
> a PR once one of the project maintainers agrees on your outlined approach.
>
> PRs of contributors who are not vouched for are automatically closed. Regular
> contributors are added to the vouch list.
>
> The kernels-community repository is for:
>
> 1. Kernels developed by Hugging Face and partners.
> 2. Kernels developed by third parties, that have not been 'kernelized' yet, but
>    are used by Hugging Face projects such as diffusers and transformers.
>
> We cannot accept PRs for the following:
>
> - New kernels that are not in categories (1) and (2).
> - Changes to kernels in category (2), these should be submitted upstream.
>
> For, LLM-generated changes, we prefer that you write your prompt in an issue
> over a PR with LLM-generated changes.

## Related issue

Closes #

## Changes

-

## Testing

-

## Checklist

- LLM disclosure:

### DIFF
--- a/finegrained-fp8/tests/test_matmul.py
+++ b/finegrained-fp8/tests/test_matmul.py
@@ -11,10 +11,9 @@ from utils import (  # type: ignore
     DTYPE_TO_TOL,
     MX_SCALE_GROUP_K,
     TEST_DEVICE,
-    accelerator_module,
-    make_fp8_weights,
-    make_fp4_weights,
     make_static_activation_scale,
+    make_weights,
+    maybe_compile,
     ref_matmul,
 )
 
@@ -27,16 +26,24 @@ class Problem:
     N: int
     K: int
     block_size: Optional[Tuple[int, int]] = None
+    weight_dtype: torch.dtype = torch.float8_e4m3fn
     weight_scale_dtype: torch.dtype = torch.float32
     static_activation_scale: bool = False
     dtype: torch.dtype = torch.bfloat16
-    weight_format: str = "fp8"
     compile: bool = False
 
     @property
     def id(self) -> str:
-        head = f"{self.weight_format}_M{self.M}_N{self.N}_K{self.K}"
-        is_mx = self.weight_format in ("mxfp4", "mxfp8")
+        # Recipe label derived from the stored dtype + scale group: packed E2M1 is
+        # MXFP4; E4M3 with a 1x32 group is MXFP8; otherwise plain FP8.
+        is_mx = self.block_size == (1, MX_SCALE_GROUP_K)
+        if self.weight_dtype == torch.int8:
+            fmt = "mxfp4"
+        elif is_mx:
+            fmt = "mxfp8"
+        else:
+            fmt = "fp8"
+        head = f"{fmt}_M{self.M}_N{self.N}_K{self.K}"
         # MX recipes ignore block_size (MX group is fixed at 32; tile autotuned).
         if is_mx:
             tail = head
@@ -94,14 +101,14 @@ PROBLEMS = [
     # group at 32 and autotunes its compute tile). ``tl.dot_scaled`` runs natively
     # on Blackwell, emulated elsewhere. ──
     Problem(
-        M=32, N=256, K=512, block_size=(1, MX_SCALE_GROUP_K), weight_format="mxfp4"
+        M=32, N=256, K=512, block_size=(1, MX_SCALE_GROUP_K), weight_dtype=torch.int8
     ),
     Problem(
         M=32,
         N=256,
         K=512,
         block_size=(1, MX_SCALE_GROUP_K),
-        weight_format="mxfp4",
+        weight_dtype=torch.int8,
         dtype=torch.float16,
     ),
     Problem(
@@ -109,7 +116,7 @@ PROBLEMS = [
         N=256,
         K=512,
         block_size=(1, MX_SCALE_GROUP_K),
-        weight_format="mxfp4",
+        weight_dtype=torch.int8,
         dtype=torch.float32,
     ),
     Problem(
@@ -117,7 +124,7 @@ PROBLEMS = [
         N=256,
         K=512,
         block_size=(1, MX_SCALE_GROUP_K),
-        weight_format="mxfp4",
+        weight_dtype=torch.int8,
         compile=True,
     ),
     # ── MXFP8 (E4M3 weights + E4M3 act, UE8M0 group-32) ──
@@ -127,7 +134,6 @@ PROBLEMS = [
         K=512,
         block_size=(1, MX_SCALE_GROUP_K),
         weight_scale_dtype=torch.float8_e8m0fnu,
-        weight_format="mxfp8",
     ),
     Problem(  # non-aligned N
         M=16,
@@ -135,7 +141,6 @@ PROBLEMS = [
         K=1024,
         block_size=(1, MX_SCALE_GROUP_K),
         weight_scale_dtype=torch.float8_e8m0fnu,
-        weight_format="mxfp8",
     ),
     Problem(
         M=32,
@@ -143,7 +148,6 @@ PROBLEMS = [
         K=512,
         block_size=(1, MX_SCALE_GROUP_K),
         weight_scale_dtype=torch.float8_e8m0fnu,
-        weight_format="mxfp8",
         dtype=torch.float16,
     ),
     Problem(
@@ -152,7 +156,6 @@ PROBLEMS = [
         K=512,
         block_size=(1, MX_SCALE_GROUP_K),
         weight_scale_dtype=torch.float8_e8m0fnu,
-        weight_format="mxfp8",
         dtype=torch.float32,
     ),
     Problem(
@@ -161,7 +164,6 @@ PROBLEMS = [
         K=512,
         block_size=(1, MX_SCALE_GROUP_K),
         weight_scale_dtype=torch.float8_e8m0fnu,
-        weight_format="mxfp8",
         compile=True,
     ),
 ]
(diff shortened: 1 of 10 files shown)


finegrained-fp8/tests/test_matmul.py
```diff
@@ -1,4 +1,4 @@
-"""Tests for ``matmul`` (FP8 and FP4 weights, including non-aligned dims)."""
+"""Tests for ``matmul`` (FP8, MXFP4, and MXFP8 weights, including non-aligned dims)."""
```
vasqu, line 1: Ah ok so prev fp4 was always mxfp4?

finegrained-fp8/tests/test_matmul.py
```diff
@@ -27,24 +26,34 @@ class Problem:
     N: int
     K: int
     block_size: Optional[Tuple[int, int]] = None
+    weight_dtype: torch.dtype = torch.float8_e4m3fn
```
vasqu, line 29: Ig this is fine because we only built for the most recent torch versions?

finegrained-fp8/tests/test_moe.py
```diff
@@ -39,23 +39,30 @@ class Problem:
...
     def id(self):
-        head = (
-            f"{self.weight_format}_S{self.S}_E{self.E}_N{self.N}_K{self.K}"
-            f"_top{self.TOP_K}"
-        )
-        # FP4 ignores block_size (tile shape autotuned); always block-mode.
-        if self.weight_format == "fp4":
+        # Recipe label derived from the stored dtype + scale group: packed E2M1 is
+        # MXFP4; E4M3 with a 1x32 group is MXFP8; otherwise plain FP8.
+        is_mx = self.block_size == (1, MX_SCALE_GROUP_K)
```
vasqu, line 55: Maybe the shared parts could be moved to a util? E.g. the id seems identical to me

finegrained-fp8/torch-ext/finegrained_fp8/__init__.py
```diff
@@ -1,20 +1,23 @@
 from .utils import fp8_act_quant
 from .batched import (
     matmul_batched,
-    w4a8_block_dynamic_fp4_matmul_batched,
```
vasqu, line 4: were we using this anywhere so far? just thinking for BC

finegrained-fp8/torch-ext/finegrained_fp8/batched.py
```diff
@@ -262,14 +264,97 @@ def w4a8_block_dynamic_fp4_matmul_batched_kernel(
...
+        a_raw = tl.load(a_ptrs).to(tl.float32)
+        a, a_scale = mx_act_quant_inline(
+            a_raw, BLOCK_SIZE_M, BLOCK_SIZE_K, SCALE_GROUP_K
+        )
+        b = tl.load(b_ptrs)
+        b_s = tl.load(bs_ptrs).to(tl.uint8)
+        accumulator = tl.dot_scaled(a, a_scale, "e4m3", b, b_s, "e4m3", acc=accumulator)
+        a_ptrs += BLOCK_SIZE_K * stride_ak
+        b_ptrs += BLOCK_SIZE_K * stride_bk
         bs_ptrs += (BLOCK_SIZE_K // SCALE_GROUP_K) * stride_bs_k
```
vasqu, line 359: Ok so we quant on the fly and then accumulate over the results. Would it make sense to create some other launch instances for those or would it be too overkill?

Just thinking that this tiling could also be parallelized but unsure whether it really is worth it

finegrained-fp8/torch-ext/finegrained_fp8/grouped.py
```diff
@@ -292,14 +295,111 @@ def w4a8_block_dynamic_fp4_matmul_grouped_kernel(
...
+    for _ in range(0, tl.cdiv(K, BLOCK_SIZE_K)):
+        a_raw = tl.load(a_ptrs, mask=row_mask[:, None], other=0.0).to(tl.float32)
+        a, a_scale = mx_act_quant_inline(
+            a_raw, BLOCK_SIZE_M, BLOCK_SIZE_K, SCALE_GROUP_K
+        )
+        b = tl.load(b_ptrs)
+        b_s = tl.load(bs_ptrs).to(tl.uint8)
+        accumulator = tl.dot_scaled(a, a_scale, "e4m3", b, b_s, "e4m3", acc=accumulator)
+        a_ptrs += BLOCK_SIZE_K * stride_ak
+        b_ptrs += BLOCK_SIZE_K * stride_bk
```
vasqu, line 497: Same question essentially here as in batched

finegrained-fp8/torch-ext/finegrained_fp8/matmul.py
```diff
@@ -317,18 +320,96 @@ def w4a8_block_dynamic_fp4_matmul_kernel(
...
+        a, a_scale = mx_act_quant_inline(
+            a_raw, BLOCK_SIZE_M, BLOCK_SIZE_K, SCALE_GROUP_K
+        )
+        b = tl.load(b_ptrs, mask=offs_k[:, None] < k_remaining, other=0.0)
+        b_s = tl.load(
+            bs_ptrs, mask=offs_sf[None, :] < k_remaining // SCALE_GROUP_K, other=0
+        ).to(tl.uint8)
+        accumulator = tl.dot_scaled(a, a_scale, "e4m3", b, b_s, "e4m3", acc=accumulator)
+        a_ptrs += BLOCK_SIZE_K * stride_ak
+        b_ptrs += BLOCK_SIZE_K * stride_bk
```
vasqu, line 414: And here

finegrained-fp8/torch-ext/finegrained_fp8/utils.py
```diff
@@ -95,38 +130,41 @@ def get_active_device_type() -> str:
...
-    on other backends — so launch sites never need to patch them in post-hoc.
+    ``with_block_sizes=True`` sweeps the tile: used by kernels that have no caller
+    ``block_size`` to fix it — the MX ``tl.dot_scaled`` paths AND the tensor-dynamic
+    FP8 paths. ``with_block_sizes=False`` emits a single empty meta-dict (block-scaled
+    kernels take the tile from the caller's ``block_size``).
+
+    The CUDA tile set is a data-driven prune of a B200 sweep across single (BM=128),
+    grouped MoE (BM=16/64) and decode (BM=1): winners only ever used these 4 tiles,
+    num_warps in {4,8,16}, num_stages in {2,3} (warps=2, stages=4 and 256x256 never
+    won) — 108 → 24. (Tuned on dot_scaled; tensor-dynamic tl.dot reuses it.)
```
vasqu, line 126: Might be worth to check out helion maybe (with triton backend) and is it automates those search grids. Maybe we could use this to automate the search space a bit

vasqu (commented):
Do we care about stability and BC for this kernel? I think we are iterating quite fast on this correct? Just wanna make sure we don't need to bump a version or similar

My overall comments are mostly small in nature. The overall structure over all implementations is similar so my comment about the tiling applies to all. 

Other than that: might be worth to try out helion? Not sure how viable it is but iirc it is automating search spaces which you do manually atm iiuc in this PR

finegrained-fp8/tests/test_matmul.py
IlyasMoutawwakil: yes, for dsv4

finegrained-fp8/torch-ext/finegrained_fp8/__init__.py
IlyasMoutawwakil: no, we just use the dispatcher, maybe we shouldn't put the recipe-specific functions in init at all, wdyt ?

vasqu: Yea seems fair to "hide" since the interface stays consistent while internals can change freely

finegrained-fp8/torch-ext/finegrained_fp8/utils.py
IlyasMoutawwakil: tried it last time to reduce the tuning time but was hard to port everything to it, at least in my experience and for my taste it doesn't let you do as much as triton (because of the abstraction)

finegrained-fp8/torch-ext/finegrained_fp8/batched.py
IlyasMoutawwakil: > Would it make sense to create some other launch instances for those or would it be too overkill?

not sure i understand 🥲

vasqu: Like a kernel in a kernel. Here we "naively" iterate over the tiles but it could be parallelized into its own kernel instead of for looping

IlyasMoutawwakil: i see, yes those partials are independent, it's a reduction over K, so it "can" be parallelized. The reason it's a loop is inside one program the accumulator lives in registers and the += is free (and the loop's already pipelined via num_stages), whereas splitting K across programs means you'd have to compute all partials, store them somewhere and reduce them globally afterwards, either with atomics (I avoid those for overhead and non-determinism) or partial buffers + a second kernel (more mem and more reads/writes). There's probably cases where that can be interesting (similar to split-k) but i don't think it is in general.

finegrained-fp8/tests/test_matmul.py
IlyasMoutawwakil: yes, and fp8 kernels can't be used without the fp8 dtype

finegrained-fp8/torch-ext/finegrained_fp8/__init__.py
IlyasMoutawwakil: then we will have to release a new version (with only public dispatches)

vasqu: I mean we are already breaking with the renames so we would need a new version either way no? Better now than later

finegrained-fp8/torch-ext/finegrained_fp8/batched.py
vasqu: Gotcha, it's a fair assessment. I was just wondering if it was worth it as it kinda sticked out. Also tbh rather keep things simple first than increasing complexity from the get go (like you said you will need new read/writes and triton is not really flexible for these kind of nested kernels so not easy to do well)

vasqu (dismissed):
Pre approving carefully because there is one failing ci it seems but should be mostly ready to go; just probably needs the version change. Thanks for sticking with my dumb questions 😬

vasqu (approved):
Thanks 🤗

finegrained-fp8/torch-ext/finegrained_fp8/__init__.py
IlyasMoutawwakil: yes agreed

PR #960, opened 2026-06-17
Add a workflow to verify all kernels

danieldk:
To be run after #958.


sayakpaul (approved):
Clean!

danieldk:
Will merge it after we removed all the old metadata builds.

PR #972, opened 2026-06-18
ci: correctly handle the case when there is nothing to sign

### LINKED ISSUE #971: Signing step fails when build produces no artifacts (glob used literally)
## Summary

When a kernel build workflow runs on a platform where there are no variants to build (e.g. running `build-cuda` for a kernel that doesn't support CUDA), the "Build kernel" step succeeds (by design — it's correct to succeed when there are no buildsets for the platform). However, the subsequent **signature verification step fails** because the glob pattern is used literally when it doesn't match any files.

## Example

- Run: https://github.com/huggingface/kernels-community/actions/runs/27025228114/job/79763455748
- "Build kernel" step succeeds but produces no artifacts (expected behavior)
- Signature verification step fails because the glob doesn't match anything and gets passed literally

## Context

- The "succeed when no variants" behavior was intentionally added so that `nix build -L .#ci` works correctly on platforms where a kernel has no buildsets.
- Before the dispatcher was added, every backend workflow would always run, and this prevented failures for unsupported platforms. Now the dispatcher avoids running inapplicable workflows, but the local build case still benefits from this behavior.

## Proposed Fix

Either:
1. Set the glob to return empty when it doesn't match anything (e.g. `nullglob` in bash), or
2. Replace the glob with `find` using the correct depth to avoid literal glob expansion when no files exist.

---
Requested by Sayak Paul (Slack U03AU4E7DJB) via Moon Bot _(no GitHub handle mapped)_.
Slack thread: https://huggingface.slack.com/archives/C090JN2P8NB/p1781766444673589?thread_ts=1781766444.673589&cid=C090JN2P8NB
Agent trace: https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-06-18T07-36-11-211Z_2745fdf4-7d05-404d-953f-a65c3b98ce8d.jsonl

danieldk:
E.g. when a workflow does not result in any builds.

Fixes #971.

### DIFF
--- a/.github/workflows/build-mac.yaml
+++ b/.github/workflows/build-mac.yaml
@@ -175,9 +175,9 @@ jobs:
         env:
           KERNEL: ${{ steps.validate.outputs.kernel }}
         run: |
-          for metadata_file in "$KERNEL"/build/*/metadata.json; do
+          while IFS= read -r metadata_file; do
             cosign sign-blob "$metadata_file" --bundle "${metadata_file}.sigstore" --yes
-          done
+          done < <(find "$KERNEL/build" -mindepth 2 -maxdepth 2 -name "metadata.json")
 
       # Upload built artifacts to both model and kernel Hub repos.
       - name: Upload kernel to Hub
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -226,9 +226,9 @@ jobs:
         if: inputs.mode != 'pr'
         run: |
           KERNEL="${{ needs.setup.outputs.kernel }}"
-          for metadata_file in "$KERNEL"/build/*/metadata.json; do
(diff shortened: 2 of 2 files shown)


.github/workflows/build-mac.yaml
```diff
@@ -175,9 +175,9 @@ jobs:
         env:
           KERNEL: ${{ steps.validate.outputs.kernel }}
         run: |
-          for metadata_file in "$KERNEL"/build/*/metadata.json; do
```
sayakpaul, line 178: For later. I think it could be nice to keep the signing utility in a script and use it from the workflows. That way, the number of places the same changes need to be made is reduced. But since the build workflows are just two at the moment, it's manageable.

sayakpaul (approved):

PR #975, opened 2026-06-18
flash-attn3: target Torch 2.9 stable ABI

danieldk:
Also try if CUDA 12.6 build succeeds.


sayakpaul:
Seems like there are some build failures.

danieldk:
> Seems like there are some build failures.

I cancelled the build because I was doing a staging build anyway, so no use tying up the runners for hours :).

vasqu (approved):
Thanks!

PR #980, opened 2026-06-19
feat: add a script to check for kernel public api changes

drbh:
this PR runs a new script to check if there are any changes to a kernels public api - when a kernels public api changes we should bump the version

### DIFF
--- a/.github/workflows/check-public-api.yaml
+++ b/.github/workflows/check-public-api.yaml
@@ -1,0 +1,34 @@
+name: Check Public API
+
+on:
+  pull_request:
+    paths-ignore:
+      - "**/*.md"
+      - ".github/**"
+      - "scripts/**"
+      - "*.json"
+      - "*.nix"
+      - "*.lock"
+      - ".gitignore"
+      - ".gitmodules"
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  check-public-api:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          fetch-depth: 0
+
+      - name: Check public API
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+        run: |
+          python3 scripts/check_public_api.py --base-ref "$BASE_SHA"
--- a/.github/workflows/check-public-api.yaml
+++ b/.github/workflows/check-public-api.yaml
@@ -27,8 +27,11 @@ jobs:
         with:
           fetch-depth: 0
 
+      - name: Install uv
+        run: pip install uv
+
       - name: Check public API
         env:
           BASE_SHA: ${{ github.event.pull_request.base.sha }}
         run: |
-          python3 scripts/check_public_api.py --base-ref "$BASE_SHA"
+          uv run --script scripts/check_public_api.py --base-ref "$BASE_SHA"
--- a/scripts/check_public_api.py
+++ b/scripts/check_public_api.py
@@ -1,0 +1,186 @@
+import argparse
+import ast
+import json
+import re
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+# Private by name but still part of the public contract.
+PUBLIC_DUNDERS = {"__init__", "__call__"}
+DEF_RE = re.compile(r"\.def\(\s*\"((?:[^\"\\]|\\.)*)\"")
+
+
+def public(name: str) -> bool:
+    return not name.startswith("_")
+
+
+def signature(node) -> str:
+    args = ast.unparse(node.args)
+    returns = f" -> {ast.unparse(node.returns)}" if node.returns else ""
+    return f"({args}){returns}"
+
+
+def class_api(node: ast.ClassDef, rel: str) -> dict:
+    out = {f"{rel} {node.name}(bases)": ", ".join(ast.unparse(b) for b in node.bases)}
+    attrs = []
+    for item in node.body:
+        if isinstance(item, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            if public(item.name) or item.name in PUBLIC_DUNDERS:
+                out[f"{rel} {node.name}.{item.name}"] = signature(item)
+        elif isinstance(item, ast.AnnAssign) and isinstance(item.target, ast.Name):
+            if public(item.target.id):
+                attrs.append(item.target.id)
+        elif isinstance(item, ast.Assign):
+            attrs += [
+                t.id for t in item.targets if isinstance(t, ast.Name) and public(t.id)
+            ]
+    if attrs:
+        out[f"{rel} {node.name}(attrs)"] = ", ".join(sorted(set(attrs)))
+    return out
+
+
+def python_api(path: Path, rel: str) -> dict:
+    out = {}
+    for node in ast.parse(path.read_text()).body:
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef)):
+            if public(node.name):
+                out[f"{rel} {node.name}"] = signature(node)
+        elif isinstance(node, ast.ClassDef) and public(node.name):
+            out.update(class_api(node, rel))
+        elif isinstance(node, ast.Assign):
+            if any(isinstance(t, ast.Name) and t.id == "__all__" for t in node.targets):
+                try:
+                    out[f"{rel} __all__"] = repr(sorted(ast.literal_eval(node.value)))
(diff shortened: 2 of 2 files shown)


scripts/check_public_api.py
```diff
@@ -0,0 +1,186 @@
...
+            if any(isinstance(t, ast.Name) and t.id == "__all__" for t in node.targets):
+                try:
+                    out[f"{rel} __all__"] = repr(sorted(ast.literal_eval(node.value)))
+                except (ValueError, TypeError):
+                    pass
+    return out
+
+
+def extract_api(kernel_root: Path) -> dict:
+    ext = kernel_root / "torch-ext"
```
sayakpaul: Okay this is only for PyTorch now. Maybe add a comment about it?

sayakpaul (dismissed):

scripts/check_public_api.py
```diff
@@ -0,0 +1,186 @@
...
+    if not ext.is_dir():
+        return {}
+    api = {}
+    for src in sorted([*ext.rglob("*.cpp"), *ext.rglob("*.cc")]):
```
danieldk: Better to read the source files from the `build.toml`?

scripts/check_public_api.py
```diff
@@ -0,0 +1,186 @@
...
+        for match in DEF_RE.finditer(src.read_text(errors="replace")):
+            schema = " ".join(match.group(1).split())
+            if schema:
+                api[f"op {schema.split('(', 1)[0].strip()}"] = schema
+    for py in sorted(ext.rglob("*.py")):
```
danieldk: The condition above skips tvm-ffi kernels, but the Python bits would apply there as well. Seems nicer to just do them separately?

Besides that, probably only stuff exported through `__all__` in `__init__.py` (when present) is public API, even more so because you cannot import other submodules. So, I think it should be checked what is reachable through `__init__.py`, otherwise we'll get a lot of false positives that are internals.

scripts/check_public_api.py
```diff
@@ -0,0 +1,186 @@
...
+
+def git(*args: str) -> str:
+    return subprocess.run(
+        ["git", *args], check=True, capture_output=True, text=True
+    ).stdout.strip()
+
+
+def changed_kernels(ref: str) -> list:
+    found = set()
+    for line in git("diff", "--name-only", ref).splitlines():
```
danieldk: Passing ref names directly to a command enables injection attacks, especially because refs can be user-controlled. Probably better to use something like libgit2?

scripts/check_public_api.py
```diff
@@ -0,0 +1,186 @@
...
+import json
+import re
+import subprocess
+import sys
+import tempfile
+from pathlib import Path
+
+# Private by name but still part of the public contract.
+PUBLIC_DUNDERS = {"__init__", "__call__"}
+DEF_RE = re.compile(r"\.def\(\s*\"((?:[^\"\\]|\\.)*)\"")
```
danieldk: Is there a C++ parser that we easily use from Python?

scripts/check_public_api.py
```diff
@@ -0,0 +1,186 @@
...
+def extract_api(kernel_root: Path) -> dict:
+    ext = kernel_root / "torch-ext"
+    if not ext.is_dir():
+        return {}
+    api = {}
+    for src in sorted([*ext.rglob("*.cpp"), *ext.rglob("*.cc")]):
+        for match in DEF_RE.finditer(src.read_text(errors="replace")):
+            schema = " ".join(match.group(1).split())
+            if schema:
+                api[f"op {schema.split('(', 1)[0].strip()}"] = schema
```
danieldk: Torch has a schema parser: `torch._C.parse_schema`.

scripts/check_public_api.py
drbh: sounds good to me, added a comment to the updated function in latest changes

scripts/check_public_api.py
drbh: yea good point - updated to use the build.toml list in latest changes

scripts/check_public_api.py
drbh: thanks for catching - the code should now inspect both `torch-ext` and `tvm-ffi-ext`'s init files now.

we either use the values from `__all__` if available, or fallback to parse all exported values that do not start with `_`

scripts/check_public_api.py
drbh: great catch, updated to prefer using `pygit2` in latest to avoid this injection risk

scripts/check_public_api.py
drbh: yes, I've updated to use libclang to parse the cpp source into tokens where the `ops.def(`  calls are identified and the signature is extracted

scripts/check_public_api.py
drbh: once the schemas are extracted in the step above they are parsed with this function in the latest changes

drbh:
PR updated and I think its much more robust/solid now.

example usage

```bash
 uv run scripts/check_public_api.py flash-attn2
```

output
```txt
Public API: 303e2f7c..71bd3f8f (origin/main..HEAD)

  [ok] flash-attn2: 18 symbols
     |- op bwd = bwd(Tensor($0! -> ) dout, Tensor($1! -> ) q, Tensor($2! -...
     |- op fwd = fwd(Tensor($0! -> ) q, Tensor k, Tensor v, Tensor(out_!)?...
     |- op fwd_kvcache = fwd_kvcache(Tensor($0! -> ) q, Tensor($1! -> ) kcache, Te...
     |- op varlen_bwd = varlen_bwd(Tensor($0! -> ) dout, Tensor($1! -> ) q, Tenso...
     |- op varlen_fwd = varlen_fwd(Tensor($0! -> ) q, Tensor k, Tensor v, Tensor?...
     |- py flash_attn2 __all__ = ['bwd', 'flash_attn_func', 'flash_attn_kvpacked_func', 'f...
     `- ... and 12 more
```

now if you add a new attribute to one of the signature (either on the cpp or python side)

```
uv run scripts/check_public_api.py flash-attn2
```

output
```txt
Public API: 303e2f7c..71bd3f8f (origin/main..HEAD)

  [changed] flash-attn2: 1 changed
     `- changed op fwd
        - fwd(Tensor($0! -> ) q, Tensor k, Tensor v, Tensor(out_!)? out_, Tensor? alibi_slopes_, float p_dropout, float softmax_scale, bool is_causal, int window_size_left, int window_size_right, float softcap, bool return_softmax, Generator? gen_) -> Tensor[]
        + fwd(Tensor! q, Tensor k, Tensor v, Tensor(out_!)? out_, Tensor? alibi_slopes_, float p_dropout, float softmax_scale, bool is_causal,int window_size_left, int window_size_right, float softcap, bool return_softmax, bool testGenerator? gen_) -> Tensor[]

ERROR: public API changed - bump the kernel version if intentional.
```


NOTE:
since we prefer parsing `__all__` and fallback to exposed functions - if a `__all__` var is defined we expect all public functions to be included. functions that are not in a defined `__all__` will not be considered as public. Therefore if a downstream user is using an exported function that is not in `__all__` it will not be caught by this check. 

TLDR;  its important to document the public api's via `__all__` in all cases.

.github/workflows/check-public-api.yaml
```diff
@@ -0,0 +1,37 @@
...
+concurrency:
+  group: ${{ github.workflow }}-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  check-public-api:
+    runs-on: ubuntu-latest
```
sayakpaul, line 24: This is such a cool way to use uv. Just rely on a Ubuntu container provided by GitHub, and then rely on uv style dependencies to run the script. Very nice!

scripts/check_public_api.py
```diff
@@ -0,0 +1,458 @@
...
+            while j < n and sp[j] != ")":
+                if sp[j] != ",":
+                    args.append(sp[j])
+                j += 1
+            if args:
+                out.append((args[0], args[1] if len(args) > 1 else args[0]))
+    return out
+
+
+# Operator schemas: PyTorch (.def, normalized via torch) and tvm-ffi exports.
```
sayakpaul, line 181: Okay it supports both frameworks. Should we handle if we added a new framework and this script doesn't know how to parse it? If we don't expect that to change for a new framework, then it's fine.

sayakpaul (approved):
Great work!

Just left a small question.

.github/workflows/check-public-api.yaml
drbh: thanks I'm a big fan of PEP 723/uv scripts!

scripts/check_public_api.py
drbh: good point, I think for now it make sense to support the two targets, we don't expect them to change frequently. If we end up adding a new framework, we can update this script then

PR #981, opened 2026-06-19
feat: Flash attn2 v2

drbh:
this PR is a follow up on https://github.com/huggingface/kernels-community/pull/976 and removes the extra files again and bumps the version since the public api is changed


danieldk:
The extra file removals seem to be missing? There is only a version bump.

danieldk (approved):

Issue #935, opened 2026-06-05
[CI] mismatch between available backend and dispatch


sayakpaul:
Currently, the bot checks out from `main` when a comment is made on a PR:
https://github.com/huggingface/kernels-community/blob/da8ff024cde741175b8afd56805e1ff9d7443a13/.github/workflows/pr-comment-build.yaml#L20

Then `dispatch.py` is called through:

https://github.com/huggingface/kernels-community/blob/da8ff024cde741175b8afd56805e1ff9d7443a13/.github/workflows/pr-comment-build.yaml#L49

https://github.com/huggingface/kernels-community/blob/da8ff024cde741175b8afd56805e1ff9d7443a13/.github/scripts/pr_comment_kernel_bot.py#L12

We have the following LOC in `dispatch.py`:

https://github.com/huggingface/kernels-community/blob/da8ff024cde741175b8afd56805e1ff9d7443a13/.github/scripts/dispatch.py#L140-L153

Since it's invoked from `main`, it means all release workflows will get triggered since backends can be none for kernels that don't exist on `main` yet:

https://github.com/huggingface/kernels-community/blob/da8ff024cde741175b8afd56805e1ff9d7443a13/.github/scripts/dispatch.py#L167-L170

I think this is what happened here:
https://github.com/huggingface/kernels-community/pull/919

The said kernel doesn't declare a "metal" backend but still, a `build-mac.yaml` was dispatched and it failed here: https://github.com/huggingface/kernels-community/actions/runs/27025228114 because it couldn't find metadata. 

My comments are:

* The dispatch should probably be done based on the content of the PR branch (now that we are making it mandatory to trigger builds through the bot after an initial vetting)? cc: @drbh 
* The build here succeeded but there was nothing produced: https://github.com/huggingface/kernels-community/actions/runs/27025228114/job/79763455748#step:12:61. Should it count valid? cc: @danieldk

arnavnaik22:
Hey @sayakpaul! I can take this one. I'll update the checkout step in the PR comment workflow to pull the actual PR branch ref instead of main, so dispatch.py can correctly read the build.toml backends from the contributor's code. Could you assign this to me?

danieldk:
I think we should dispatch from `main` because PRs may not have rebased and the dispatch script may have newer changes. However, the metadata should be read from the branch. So it seems like the dispatch should have some additional logic.

[...]

PR #982, opened 2026-06-19
feat: prefer reading the build toml from current ref

drbh:
this PR updates the kernel bot to prefer reading the build toml from the PR ref its on

pending the refactor PR https://github.com/huggingface/kernels-community/pull/983

should address https://github.com/huggingface/kernels-community/issues/935

### DIFF
--- a/.github/scripts/test_dispatch.py
+++ b/.github/scripts/test_dispatch.py
@@ -1,5 +1,6 @@
 import io
 import os
+import subprocess
 import sys
 import unittest
 import urllib.error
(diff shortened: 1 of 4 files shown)


sayakpaul:
Cc: @drbh possible to revive this one?

.github/scripts/test_dispatch.py
```diff
@@ -68,6 +69,31 @@ def test_security_only_plans_only_security(self):
             self.assertEqual(action.body["inputs"]["head_sha"], "deadbeef")
 
 
+# Metadata read from a git ref (PR branch) instead of the working tree.
+class MetadataRefTest(unittest.TestCase):
```
sayakpaul, line 73: I think there should also be a test to ensure that the dispatcher is able to select the right workflows? For example, https://github.com/huggingface/kernels-community/pull/1031 shouldn't have picked the building on Mac workflow?

drbh: yea good point - `test_metadata_ref_routes_metal_less_kernel_without_mac_build` should cover this case, it reads a cuda only kernel from the ref and asserts it only routes to `build.yaml` (no `build-mac.yaml`)

thats what happened in #1031, `build.toml` didn't exist on the base ref yet so `read_backends` returned `None` and we fell back to dispatching every build, mac included. reading the toml from the PR ref is the fix

sayakpaul (approved):
Thanks! Let's ship it and fix any bugs that might surface later.

PR #983, opened 2026-06-19
feat: refactor kernel bot and add tests

drbh:
this PR refactors the kernel bot dispatch to split planning from execution, so dry runs don't hit the api and the logic is actually testable.

- split the work into `plan` and `execute`
- `plan_dispatch()` is pure: it reads build.toml, selects workflows, and builds the dispatch payloads/keys with no network calls, token, or repo.
- `execute_plan()` performs I/O, workflow dispatches and pending commit statuses.
- consolidates build and security dispatches
- `pr_comment_kernel_bot.py` now supports a `--dry-run` flag to allow testing locally
- added tests for dispatch and pr_comment_kernel_bot to avoid future regressions

run tests
```bash
python3 -m unittest discover -s .github/scripts -p "test_*.py"
```

using dry run for comment tests
```bash
python .github/scripts/pr_comment_kernel_bot.py --dry-run "/kernel-bot security" --pr-number 982
```

using dry run on the dispatch
```bash
python .github/scripts/dispatch.py --security-only --pr-number 982 --head-sha "$(gh pr view 982 --json headRefOid -q .headRefOid)" --dry-run
```

### DIFF
--- a/.github/scripts/dispatch.py
+++ b/.github/scripts/dispatch.py
@@ -1,14 +1,3 @@
-#!/usr/bin/env python3
-"""
-Dispatch build workflows for a kernel.
-
-Four entrypoints call this script:
-  1. The PR-merge dispatch workflow (via CLI)
-  2. The PR-open dispatch workflow (via CLI)
-  3. The comment bot (via import)
-  4. Local CLI invocation
-"""
-
 import argparse
 import json
 import os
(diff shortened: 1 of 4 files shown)


.github/scripts/dispatch.py
```diff
@@ -23,92 +12,22 @@
...
-        if match:
-            return match.group(1)
-    except (FileNotFoundError, subprocess.CalledProcessError):
-        pass
-    return None
-
-
-# Maps workflow filename to the `name:` field in the YAML, used for commit
-# status context strings that must match between pending and final statuses.
+# Workflow file -> its YAML `name:`; must match the context the build reports.
```
sayakpaul, line 30: What does it mean? 👀

.github/scripts/dispatch.py
```diff
@@ -123,22 +42,60 @@ def get_repo() -> str | None:
...
     # "flash-attn2",
 }
 
-# Backends to skip on Windows for specific kernels (e.g. due to toolchain issues).
 WINDOWS_SKIP_BACKENDS: dict[str, set[str]] = {
     "flash-attn2": {"xpu"},  # CUTLASS XPU headers fail on Windows (ushort undefined)
 }
 
 
+# Data types
```
sayakpaul, line 61: Super nice!

.github/scripts/dispatch.py
```diff
@@ -152,81 +109,342 @@ def read_backends(kernel_name: str) -> list[str] | None:
...
     workflows = set()
     for b in backends:
         workflows.update(BACKEND_TO_WORKFLOWS.get(b, set()))
 
     if not workflows:
-        print(f"No known backends found for {kernel_name}: {backends}, dispatching all workflows")
-        return set(RELEASE_WORKFLOWS)
+        notes.append(
+            f"No known backends found for {kernel_name}: {backends}, dispatching all workflows"
+        )
```
sayakpaul, line 127: Curious -- when would this happen?

.github/scripts/dispatch.py
```diff
@@ -152,81 +109,342 @@ def read_backends(kernel_name: str) -> list[str] | None:
...
+    if head_sha:
+        inputs["head_sha"] = head_sha
+    if target_branch:
+        inputs["target_branch"] = target_branch
+    if not upload:
+        inputs["upload"] = "false"
+    return inputs
+
+
+def _plan_security_actions(
```
sayakpaul, line 175: Do we need planning for security, though?

.github/scripts/test_dispatch.py
```diff
@@ -0,0 +1,211 @@
...
+        self.assertEqual(self._select(None), set(dispatch.WORKFLOWS["build"]))
+
+    def test_unknown_backends_falls_back_to_all_builds(self):
+        self.assertEqual(
+            self._select(["totally-made-up"]), set(dispatch.WORKFLOWS["build"])
+        )
+
+
+# Pure planning: backends and flags -> the intended dispatch actions.
+class PlanDispatchTest(unittest.TestCase):
```
sayakpaul: (nit): let's use `pytest`.

.github/scripts/dispatch.py
drbh: this is just a small mapping between the yaml file an the `name` value that the workflow defines

we need this name since when we post status updates in the CI tracker, the github api expects the workflows name (not its file name). 

specifically/technically this value gets set as the `PlannedDispatch.status_context` which is later used in `_set_pending_status` to update the status for the given workflow

.github/scripts/dispatch.py
drbh: this should only happen if there is no backend set or if the backends set to not match any of the known backends (I guess in the case there are typos or etc)

I think this case is unlikely - but let me know if I should update the default functionality!

.github/scripts/dispatch.py
drbh: ehh kinda, one of the major changes in this PR is that we separated the "planning" step from the "execute" step.  

planning in this case builds a list of all of the api calls that need to be executed. 

for the build scripts there is a bit of extra logic to determine which builds to fire off - security doesn't have this requirement - but it's still "planned" since its placed into the list of things that need to be executed. 

the reason for this separation, is to make it easier to debug/test changes to the "planned" actions. when adding more features/complexity we can make changes to the planning logic to see how that impacts the plan without having to execute the api calls.

.github/scripts/test_dispatch.py
drbh: ah yea that makes sense, updated to use pytest in latest changes


```bash
uv run python -m pytest .github/scripts/test_dispatch.py .github/scripts/test_pr_comment_kernel_bot.py
```

```txt
============================= test session starts =============================
platform darwin -- Python 3.13.5, pytest-9.0.3, pluggy-1.6.0
rootdir: /Users/drbh/Projects/kernels-community
plugins: anyio-4.13.0
collected 36 items

.github/scripts/test_dispatch.py ..................                     [ 50%]
.github/scripts/test_pr_comment_kernel_bot.py ..................        [100%]

============================= 36 passed in 0.17s ==============================
```

.github/scripts/test_dispatch.py
```diff
@@ -0,0 +1,199 @@
+import io
```
danieldk, line 1: Nice that this can be properly tested now!

danieldk (approved):

PR #989, opened 2026-06-23
vllm-moe: welcome back to kernels-community

danieldk:
This imports vllm-moe into kernels-community again. This is the original version, updated to work with the latest kernel-builder. Syncing with vLLM will be done in a follow-up PR.


scripts/check_kernel_freshness.py
```diff
@@ -55,7 +55,8 @@
     "bitsandbytes-mps": "",
     "aiter-flash-attn": "https://github.com/ROCm/aiter",
     "aiter-kernels": "https://github.com/ROCm/aiter",
-
+    # Too much stuff happening in vLLM, not worth reporting as MoE freshness.
```
sayakpaul, line 58: Yeah 💯

sayakpaul (approved):
Superb work!

PR #992, opened 2026-06-23
aiter-flash-attn: rename public sink param to s_aux

Abdennacer-Badaoui:
transformers' FA introspection only recognizes `s_aux` (vllm-fa3) or `learnable_sink` (FA4) as the sink-param name. Our public arg was `sink`, so transformers silently dropped the sink tensor on the way to the kernel and attention ran without sinks (visible as degenerate repetitive output on gpt-oss). Renaming the public param to `s_aux` (internal kernel path unchanged) fixes the wire-up.

### DIFF
--- a/aiter-flash-attn/torch-ext/aiter_flash_attn/mha.py
+++ b/aiter-flash-attn/torch-ext/aiter_flash_attn/mha.py
@@ -515,7 +515,7 @@ def flash_attn_func(
     deterministic=True,
     return_lse=False,
     return_attn_probs=False,
-    sink=None,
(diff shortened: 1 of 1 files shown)


vasqu (dismissed):
Good from my side not sure whether we raise versions here? cc @danieldk 

Is versioning only related to `layers` or anything that faces the public API in some way?

Abdennacer-Badaoui:
No need for this. We will allow multiple different names for the same argument here : https://github.com/huggingface/transformers/pull/45153

PR #994, opened 2026-06-24
mamba-ssm: vendor causal_conv1d to remove dependency

danieldk:


vasqu:
Can you ping me when I can use from staging or similar? Would love to try this locally and make suggestions based on that :saluting_face:

mamba-ssm/torch-ext/mamba_ssm/_causal_conv1d/causal_conv1d_interface.py
vasqu, line 1: Could we also expose these as layers.py

```py
from ._causal_conv1d import causal_conv1d_fn as cuda_causal_conv1d_fn
from ._causal_conv1d import causal_conv1d_update as cuda_causal_conv1d_update


def causal_conv1d_fn(
    hidden_states,
    weight,
    bias=None,
    activation=None,
    **kwargs,
):
    # For varlen
    seq_idx = kwargs.pop("seq_idx", None)

    return cuda_causal_conv1d_fn(
        x=hidden_states,
        weight=weight,
        bias=bias,
        activation=activation,
        seq_idx=seq_idx,
    )


def causal_conv1d_update(
    hidden_states,
    conv_state,
    weight,
    bias=None,
    activation=None,
):
    return cuda_causal_conv1d_update(
        x=hidden_states,
        conv_state=conv_state,
        weight=weight,
        bias=bias,
        activation=activation,
    )


__all__ = ["causal_conv1d_fn", "causal_conv1d_update"]
```

vasqu (dismissed):
I was mostly converned for the causal conv1d. Imo it would be nice to load and expose the conv1d stuff as well

E,g, in an init
```py
__version__ = "2.2.4"

from . import layers
from .layers import causal_conv1d_fn, causal_conv1d_update
from .models import MambaLMHeadModel
from .modules import Mamba, Mamba2
from .ops import mamba_chunk_scan_combined, mamba_split_conv1d_scan_combined, selective_state_update
from .ops.selective_scan_interface import mamba_inner_fn, selective_scan_fn


falcon_mamba_inner_fn = mamba_inner_fn
__all__ = [
    # wrappers
    "layers",
    "causal_conv1d_fn", "causal_conv1d_update",
    # originals
    "selective_scan_fn", "mamba_inner_fn", "falcon_mamba_inner_fn", "selective_state_update", "mamba_chunk_scan_combined", "mamba_split_conv1d_scan_combined",
    "Mamba", "Mamba2", "MambaLMHeadModel",
]
```

It would help a lot and also verified locally that it works

mamba-ssm/torch-ext/mamba_ssm/_causal_conv1d/causal_conv1d_interface.py
sayakpaul: Is that how we're supposed to expose layers? I thought they're supposed to follow the blueprint of https://huggingface.co/kernels/kernels-community/layer-norm/blob/v1/build/torch210-cxx11-cu126-aarch64-linux/layers.py ?

vasqu: It depends on whether it's a function or a normal nn.Module with a forward - at least in old kernels we have to expose them on init as well. But I'm not sure if we make the jump to bump to 0.16.0 in transformers before. 

Maybe we can also adapt to support both from the get go - for older and newer kernels?

sayakpaul: I think newer is better, as it helps reduce the maintenance burden. But if it's an absolute must, then we could consider.

sayakpaul: I would also like to take @drbh's opinion here.

vasqu: I mean it's hard to tell because the transformers bump is currently blocked so I'm kind of forced to use older variations until then 😓

sayakpaul: I think you're meaning to expose those functions in layers? 👀 If so, would you be able to do a PR to this branch?

vasqu: Will do!

vasqu: #1019

vasqu (approved):

PR #995, opened 2026-06-25
liger-kernels: Add layer repos for remaining functions

vasqu:
As per title, adds layer type versions for the functions


liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -252,6 +252,82 @@ def LigerForCausalLMLoss(
 LigerForCausalLMLoss.can_torch_compile = True
 
 
+class liger_rotary_pos_emb_layer(nn.Module):
```
vasqu, line 255: added layer suffices as it would clash with the original fns otherwise and we should keep BC imo

liger-kernels/torch-ext/liger_kernels/layers.py
```diff
@@ -252,6 +252,82 @@ def LigerForCausalLMLoss(
...
+        return LigerRopeFunction.apply(q, k, cos, sin, None, unsqueeze_dim)
+
+
+class LigerForCausalLMLossLayer(nn.Module):
+    can_torch_compile = True
+
+    # NOTE: We have this intentional graph break as we encounter issues such as IMAs and Cublas errors.
+    #       We know that this is an optimized kernel already so there is less ways to
+    #       fuse it either way; we rely on torch compile to go through the base model to optimize.
+    @torch.compiler.disable
```
sayakpaul, line 276: Where does it fail?

vasqu: It would need a proper fake wrapping but we already do the same for the function. I would really like to move this to a separate PR / fix this upstream in liger kernels directly

drbh (dismissed):
lgtm

drbh (approved):

PR #996, opened 2026-06-25
aiter-rope: Add layer repo for apply_rotary_transformers

Abdennacer-Badaoui:
Adds an `nn.Module` layer version of `apply_rotary_transformers` (mirrors #949 for rotary) so it works with the layer-repo path in huggingface/transformers#46883. Function kept for BC.


drbh (approved):
lgtm, but just kicked off a build to make sure it builds

sayakpaul:
The dispatch failure is a fault on our end I think. So, starting a build manually.

PR #997, opened 2026-06-25
feat: prefer torch stable abi for cuda fa2

drbh:
work in progress to move fa2 to torch stable abi

>[!WARNING]
> stable abi 2.10 does not support RNG so dropout is disabled.

related PR https://github.com/Dao-AILab/flash-attention/issues/2691


flash-attn2/build.toml
```diff
@@ -1,7 +1,8 @@
 [general]
 name = "flash-attn2"
-version = 2
+version = 3
```
sayakpaul, line 3: Are we changing signatures?

flash-attn2/build.toml
```diff
@@ -1,7 +1,8 @@
...
 license = "BSD-3-Clause"
+edition = 5
```
sayakpaul, line 5: Remind me why do we need the `edition` field? Didn't find too much explanation under https://huggingface.co/docs/kernels/main/en/builder/writing-kernels#general.

sayakpaul (dismissed):
Super cool!

flash-attn2/torch-ext/torch_binding_stable.cpp
```diff
@@ -0,0 +1,137 @@
...
+    "int max_seqlen_q, "
+    "int max_seqlen_k, "
+    "float p_dropout, float softmax_scale, "
+    "bool zero_tensors, "
+    "bool is_causal, "
+    "int window_size_left, "
+    "int window_size_right, "
+    "float softcap, "
+    "bool deterministic, "
+    "Generator? gen_, "
```
janeyx99, line 101: I don't think Generator is supported in the schema as of 2.10--I added it in 2.13! But I suspect this works because gen_ is always None? Could you call out the supported/no longer supported features in this implementation? I've also been trying to get upstream to adopt stable dropoutless FA2 through a more explicit approach: https://github.com/Dao-AILab/flash-attention/issues/2691

flash-attn2/build.toml
drbh: yes we ignore an argument and drop support for dropout

flash-attn2/build.toml
drbh: this was just added for backend scoped stable abis, its basically a versioning for the metadata, an earlier discussion is here https://github.com/huggingface/kernels/pull/682#discussion_r3511474547

flash-attn2/torch-ext/torch_binding_stable.cpp
drbh: ah thanks for pointing out, yea the Generator is always unused in these changes and was silently dropped. I've updated to throw if a generator is passed to be more explicit about not supporting.

and I've updated the pr description to more clearly indicate that dropout is disabled in this change set.

also longer term it would be great to replace the current approach with your changes once they are upstreamed. I'll keep and eye for the merge and circle back! 

thanks for the help @janeyx99!

sayakpaul (approved):

sayakpaul:
Cc: @vasqu regarding https://huggingface.slack.com/archives/C090JN2P8NB/p1785356621676839

PR #998, opened 2026-06-26
vllm-flash-attn3: switch to Torch stable ABI

danieldk:
- Switch to Torch stable ABI.
- Remove duplicated builds of the same kernel files.
- Add a build configuration for quick testing.
- Try to build for CUDA 12.6 as well.
- Switch tests to use `get_kernel`.


vllm-flash-attn3/build.toml.test
```diff
@@ -0,0 +1,275 @@
+# Copy this file to `build.toml` for testing the build without
+# waiting for hours. Produces a very minimal, no-frills version
```
sayakpaul, line 2: Example ops that we skip here? Might be worth documenting for future references.

vllm-flash-attn3/flake.nix
```diff
@@ -13,5 +13,12 @@
     kernel-builder.lib.genKernelFlakeOutputs {
       inherit self;
       path = ./.;
+
+      torchVersions =
+        allVersions:
+        builtins.map (
+          version:
+          if (version.cudaVersion or null) == "12.6" then version // { ptxasVersion = "12.8"; } else version
+        ) allVersions;
```
sayakpaul, line 23: Worth adding a comment on why we needed it.

sayakpaul (dismissed):

PR #1000, opened 2026-06-29
Fix tiled MLP for ZeRO-3 and FSDP

kashif:
The tiled MLP kernel passes `compute_params` through but never uses it, and computes `num_shards` locally on each rank. Both bite under sharded-parameter backends:

- DeepSpeed ZeRO-3 crashes: the per-shard recompute fires ZeRO's gradient hook multiple times for the same parameter, so the same `ds_id` lands in the reduction bucket twice and it asserts.
- ZeRO-3 / FSDP can deadlock: if ranks see different sequence lengths they pick different `num_shards`, so they issue a different number of weight all-gathers and hang.

This ports the fixes that DeepSpeed's own TiledMLP uses:

- In the backward, defer ZeRO-3 reduction by setting `param.ds_grad_is_ready = False` for every shard except the last (gated on the `ds_id` marker, so non-ZeRO backends are untouched).
- In `apply_tiled_mlp`, all-reduce `num_shards` to the per-rank max when distributed, so every rank runs the same number of shards.

Same fix is going into linkedin/Liger-Kernel. Verified there on 2x H100: tiled gradients match a non-tiled reference under FSDP2 and DeepSpeed ZeRO-3 (with and without gradient checkpointing). No change for single-GPU or non-distributed runs.

### DIFF
--- a/liger-kernels/torch-ext/liger_kernels/tiled_mlp.py
+++ b/liger-kernels/torch-ext/liger_kernels/tiled_mlp.py
@@ -5,6 +5,7 @@ from typing import List
 from typing import Optional
 
 import torch
+import torch.distributed as dist
 
 from .utils import ensure_contiguous
 
@@ -86,6 +87,10 @@ class LigerTiledMLPFunction(torch.autograd.Function):
         x_shards = list(torch.chunk(x, chunks=shards, dim=0))
         incoming_grad_shards = list(torch.chunk(incoming_grad, chunks=shards, dim=0))
 
+        # ZeRO-3 partitioned parameters carry a ds_id; collect them once so the per-shard loop only
+        # flips the ready flag. Parameters on other backends have no ds_id and are left untouched.
+        ds_params = [p for p in compute_params if hasattr(p, "ds_id")] if compute_params else []
+
         for i, x_shard in enumerate(x_shards):
             x_shard.requires_grad_(x_requires_grad)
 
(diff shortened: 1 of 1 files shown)


liger-kernels/torch-ext/liger_kernels/tiled_mlp.py
```diff
@@ -94,6 +96,15 @@ def backward(ctx, *grads) -> tuple:
             x_shard.grad = x_grad.narrow(0, shard_offset, shard_step).view_as(x_shard)
             incoming_grad_shard = incoming_grad_shards[i]
 
+            # Under ZeRO-3 each shard's recompute would queue the same parameter for gradient reduction,
+            # so defer DeepSpeed's reduction until the last shard has accumulated into param.grad. The
+            # ds_id attribute marks a ZeRO-3 partitioned parameter, leaving other backends untouched.
+            if compute_params:
```
vasqu: Can be moved outside of the for loop no? Or does `ds_id` interfere somehow?

liger-kernels/torch-ext/liger_kernels/tiled_mlp.py
```diff
@@ -130,6 +141,15 @@ def apply_tiled_mlp(
     # Ensure num_shards is at least 1
     num_shards = max(1, num_shards)
 
+    # All ranks must run the same number of shards: a sharded-parameter backend (DeepSpeed ZeRO-3, FSDP)
+    # gathers weights inside each shard's recompute, so a rank that runs fewer shards stops participating
+    # in those collectives and deadlocks the others. Harmonize on the per-rank maximum.
+    dist = torch.distributed
```
vasqu: nit would move the import to top file

vasqu (approved):
Thanks 🤗

danieldk:
And congrats on making PR #1000 @kashif :)

kashif:
thanks @danieldk happy to help!

PR #1007, opened 2026-06-30
feat: activation stable abi

drbh:
this PR adds the stable torch ABI to the activation kernel and also exercises the backend scoped stable abi feature that is implemented in this branch https://github.com/huggingface/kernels/pull/682

### DIFF
--- a/activation/activation/dispatch_utils.h
+++ b/activation/activation/dispatch_utils.h
@@ -1,36 +1,173 @@
 /*
- * Stable-ABI dtype dispatch for the activation kernels.
+ * Stable ABI compatible dispatch utilities for vLLM.
+ * Adapted from dispatch_utils.h to use PyTorch's header-only (THO_*) macros
+ * instead of the ATen (AT_*) macros.
  *
- * The non-stable build used ATen's AT_DISPATCH_* machinery (from
- * <torch/all.h>), which is not part of Torch's stable ABI. Here we provide a
- * minimal, header-only replacement that switches on the runtime dtype and
- * binds `scalar_t` to the corresponding header-only scalar type before
- * invoking the provided lambda.
+ * These macros use:
+ * - THO_DISPATCH_SWITCH instead of AT_DISPATCH_SWITCH
+ * - THO_DISPATCH_CASE instead of AT_DISPATCH_CASE
+ * - torch::headeronly::ScalarType instead of at::ScalarType
+ *
+ * Add more macros here as needed when migrating additional kernels.
  */
 #pragma once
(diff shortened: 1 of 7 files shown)


activation/activation/dispatch_utils.h
```diff
@@ -1,83 +1,36 @@
...
+    switch (TYPE) {                                                           \
+      VLLM_DISPATCH_CASE_FLOATING_TYPE(Float, float, __VA_ARGS__)             \
+      VLLM_DISPATCH_CASE_FLOATING_TYPE(Half, torch::headeronly::Half,         \
+                                       __VA_ARGS__)                           \
+      VLLM_DISPATCH_CASE_FLOATING_TYPE(BFloat16, torch::headeronly::BFloat16, \
+                                       __VA_ARGS__)                           \
+      default:                                                                \
+        STD_TORCH_CHECK(false, NAME " not implemented for the given dtype");  \
+    }                                                                         \
+  }()
```
danieldk: I think this removes float8?

We should sync this file against: https://github.com/vllm-project/vllm/blob/main/csrc/libtorch_stable/dispatch_utils.h

activation/build.toml
```diff
@@ -16,6 +16,9 @@ src = [
     "torch-ext/torch_binding.h",
 ]
 
+[torch.stable-abi]
+cuda = "2.11"
```
danieldk: ROCm/HIP?

Also, is there a reason that 2.9 cannot be targeted?

drbh: rocm was originally left out because this kernel was only targeting cuda and metal, updated to also target rocm in latest.

and I think that the vllm dispatch header requires us to use 2.10 and later because of `torch/headeronly/core/Dispatch.h`

it seems like that files only exists in 2.10 and later
https://github.com/pytorch/pytorch/blob/v2.10.0/torch/headeronly/core/Dispatch.h
https://github.com/pytorch/pytorch/blob/v2.9.0/torch/headeronly/core/Dispatch.h

I think we can either unsync the dispatch header and target 2.9 or sync with vllm and target 2.10

activation/activation/dispatch_utils.h
drbh: ah yea good point it is much better to sync wth that upstream file, I originally slimmed it down because it wasn't clear to me that fp8 was used. 

updated to sync with vllm in latest!

danieldk (dismissed):
Looks good!

sayakpaul (approved):
Approving since @danieldk already approved before.

PR #1010, opened 2026-07-02
fla: add kdn kernels

vasqu:
As per title, starts to add the kernels from FLA. Since these are mostly implemented in triton there is no extra deps for the most part. There are some tilelang implementations but would ignore them for now. Same kernels also over here for example: https://huggingface.co/kernels/AntonV/fla

I only focused on gdn and kdn but would like to gradually update this. Should this be version 0 then maybe.


fla/torch-ext/fla/layers.py
```diff
@@ -0,0 +1,21 @@
+import torch.nn as nn
+
+from .modules.fused_norm_gate import rms_norm_gated
+
+
+class FusedRMSNormGated(nn.Module):
```
vasqu, line 7: This is for the future, e.g. we can use this for qwen 3.5

sayakpaul: If it's for the future, does it need to be present in this PR?

fla/build.toml
```diff
@@ -0,0 +1,15 @@
...
+name = "fla"
+version = 1
+license = "MIT"
+backends = [
+  "cuda",
+]
+python-depends = ["einops"]
+
+[general.hub]
+repo-id = "kernels-community/fla"
```
sayakpaul, line 11: What do we think about asking the partner if we could host the kernels under their namespace, similar to what we did for Minimax? Cc: @LysandreJik 

For the build, we can push to `kernels-community` and then have one of the admins transfer it to their namespace. But after that, we can update this to the partner namespace (once there is agreement).

scripts/check_kernel_freshness.py
```diff
@@ -58,6 +58,7 @@
     "aiter-kernels": "https://github.com/ROCm/aiter",
     # Too much stuff happening in vLLM, not worth reporting as MoE freshness.
     "vllm-moe": "",
+    "fla": "github.com/fla-org/flash-linear-attention",
```
sayakpaul, line 62: Sweet!

sayakpaul (dismissed):
Let's get the nix formatting fixed.

fla/torch-ext/fla/layers.py
vasqu: Yes, confirmed that I will need this internally. I will check a bit if this is the correct variation before merge tho

fla/build.toml
vasqu: So you mean under the FLA organization? Imo, that sounds good to me; wanted to credit them in shape or form in any case

sayakpaul: Yup. So something like `FLA/fla` or whatever their org namespace is.

vasqu:
Checking whether to add causal conv1d here as well so withholding for a bit until I've settled on that

sayakpaul (dismissed):
Thanks. Let's try to transfer the kernel repo to the actual org if possible after it's been merged.

We're working on https://github.com/huggingface/kernels/pull/694 which will allow us to open PRs to these repos (where we don't have write access) through our CI automatically :)

sayakpaul (approved):
Let's wait till the build is successful?

PR #1011, opened 2026-07-03
ci: security hardening

danieldk:
- Do not directly interpolate variables used in shell-based executions.
- Restrict Claude to local, read-only tools.
- Proactively verify `KERNEL` before it is used.

### DIFF
--- a/.github/workflows/benchmark.yaml
+++ b/.github/workflows/benchmark.yaml
@@ -44,10 +44,25 @@ jobs:
             fi
             echo "kernel=$KERNEL" >> $GITHUB_OUTPUT
           fi
+
+      # Reject kernel names outside the safe charset before they are used in Hub
+      # repo paths (guards against path traversal / repo retargeting).
+      - name: Validate kernel name
+        if: steps.kernel.outputs.skip != 'true'
+        env:
+          KERNEL: ${{ steps.kernel.outputs.kernel }}
+        run: |
+          case "$KERNEL" in
+            ''|*[!A-Za-z0-9_-]*)
+              echo "Invalid kernel name '$KERNEL': must match ^[A-Za-z0-9_-]+$"
+              exit 1
+              ;;
+          esac
+
       - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405 # v6.2.0
         if: steps.kernel.outputs.skip != 'true'
         with:
-          python-version: '3.12'
+          python-version: "3.12"
       - name: Install kernels
         if: steps.kernel.outputs.skip != 'true'
         run: pip install "kernels[benchmark] @ git+https://github.com/huggingface/kernels#subdirectory=kernels"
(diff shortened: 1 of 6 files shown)


.github/workflows/benchmark.yaml
```diff
@@ -44,10 +44,25 @@ jobs:
             fi
             echo "kernel=$KERNEL" >> $GITHUB_OUTPUT
           fi
+
+      # Reject kernel names outside the safe charset before they are used in Hub
+      # repo paths (guards against path traversal / repo retargeting).
+      - name: Validate kernel name
```
sayakpaul, line 51: Maybe an example of a valid kernel name and an invalid one?

.github/workflows/generate-build-matrix.py
```diff
@@ -36,6 +38,25 @@ def load_concurrency_overrides() -> dict:
...
+        )
+        sys.exit(1)
+    return backend
+
+
+def validate_int(value, name: str) -> int:
+    # bool is a subclass of int, but we don't want to accept it either.
+    if isinstance(value, bool) or not isinstance(value, int):
+        print(f"Invalid {name} value: {value!r} (expected integer)", file=sys.stderr)
+        sys.exit(1)
```
sayakpaul, line 56: Can't we raise here? Why are we exiting?

.github/workflows/security-audit.yml
```diff
@@ -215,7 +217,13 @@ jobs:
           === DIFF ===
           PROMPT
             cat /tmp/changes.diff
-          } | claude -p --model claude-opus-4-8 > /tmp/audit_result.txt
+          } | claude -p \
+                --model claude-opus-4-8 \
+                --permission-mode plan \
+                --allowedTools "Read,Grep,Glob,LS" \
```
sayakpaul, line 223: WebFetch and WebSearch don't look vulnerable and it might be helpful for the agent to do websearch to gain understanding about particular elements in a diff?

.github/workflows/generate-build-matrix.py
danieldk: I think a short print here is nicer than an exception. It gives full information and you do not have to go through the noisy output of an exception.

.github/workflows/security-audit.yml
danieldk: WebFetch enables secret exfiltration, the docs also recommend against this:

> Enabling the web fetch tool in environments where Claude processes untrusted input alongside sensitive data poses data exfiltration risks. Only use this tool in trusted environments or when handling non-sensitive data.

https://platform.claude.com/docs/en/agents-and-tools/tool-use/web-fetch-tool

I think `WebSearch` is safer, but an attacker could use it to add additional instructions that are not visible as part of the PR itself.

.github/workflows/benchmark.yaml
danieldk: Added.

sayakpaul (approved):

PR #1013, opened 2026-07-04
Add RDNA3.5 (gfx1150/gfx1151) configs for aiter-flash-attn

kashif:
On RDNA3.5 the kernel had no MHA tuning config, so `_get_config` hit a `FileNotFoundError` (`gfx1150-MHA-DEFAULT.json`) on the first attention call.

gfx1151 (Strix Halo) is taken straight from upstream aiter. gfx1150 (Strix Point) reuses it since both are RDNA3.5.

Tested on a gfx1150: Triton compiles the kernel fine and the forward output matches the SDPA baseline in a small transformers training run.


Abdennacer-Badaoui (approved):
LGTM! Thanks @kashif

sayakpaul (commented):
@Abdennacer-Badaoui should we not update the `build.toml` for this?

Abdennacer-Badaoui:
No, i don't think so. there's nothing to change.

sayakpaul (approved):

PR #1014, opened 2026-07-05
finegrained-fp8: add GRF register-pressure pruner for XPU autotune

sywangyi:
Add warp_size() helper (reads from Triton's active driver, falls back to max(sub_group_sizes) then 32) and grf_config_pruner() — an early_config_prune that estimates per-thread register usage and drops configs whose GRF pressure would trigger the Intel large-GRF double-compile cycle. Even after recompile, these configs can leave tens of thousands of residual spills and never win the tune; pruning them up front eliminates the wasted compile + bench time.

@IlyasMoutawwakil


sywangyi:
take python -m pytest tests/test_moe.py -k u8scale -v -s for example
without optimization, it takes about 32min to finish the 2 cases
with optimization, it takes about 8min to finish the 2 cases


pytest tests/test_moe.py -k "fp8_b128x128" -v -s
without optimization, it takes about 10min to finish the 6 cases
with optimization, it takes about 1min to finish the 6 cases

sywangyi:
for fused_grouped optimization using device_descriptor load paths and transposed accumulator layouts

This layout improves XPU DPAS utilization for grouped MoE workloads where the per-expert M tile is often small or sparse. The explicit pointer-load paths remain as fallback for non-device_descriptor modes.

Measured locally, this provides roughly 2-4x speedup on the tested W8A8 fused grouped MoE shapes:
fp8_b128x128_T128_E128_H2048_I768_top8_bf16
fp8_b128x128_T32_E128_H2048_I768_top8_bf16
fp8_b128x128_T8_E8_H512_I256_top8_bf16_sentinel
fp8_b128x128_T4_E8_H512_I256_top8_bf16
fp8_b128x128_T1_E8_H512_I256_top8_bf16

sywangyi:
I add the corresponding skill in https://github.com/huggingface/kernels/pull/690

finegrained-fp8/tests/utils.py
```diff
@@ -133,7 +133,10 @@ def make_weights(
         )
         qmax = FP8_MAX
 
-    W = torch.randn(E, out_features, in_features, dtype=torch.float32, device=device)
+    # Build on CPU to avoid large float32 peak allocation on the accelerator.
+    # For large MoE shapes (e.g. E=256, H=4096) the fp32 scratch tensor can reach
+    # 16 GiB; quantized outputs are moved to ``device`` only at the very end.
+    W = torch.randn(E, out_features, in_features, dtype=torch.float32)
```
IlyasMoutawwakil: this will take a lot of time tho .. cuda random number generations is orders of magnitude faster than on cpu

finegrained-fp8/torch-ext/finegrained_fp8/fused_grouped.py
```diff
@@ -213,14 +215,21 @@ def _grouped_routing(expert_ids: torch.Tensor, num_experts: int, num_top_k: int)
 
 # ── Block-dynamic FP8 ────────────────────────────────────────────────────────
 
-
+_W8A8_MEMORY_MODES = ("device_descriptor", "block_ptr")
```
IlyasMoutawwakil: this hardcodes device descriptors (we need host on cuda devices)

IlyasMoutawwakil:
the config pruner looks great ! not sure about the other changes as the intention is not documented

finegrained-fp8/torch-ext/finegrained_fp8/fused_grouped.py
sywangyi: _W8A8_MEMORY_MODES is currently used in w8a8_block_dynamic_fp8_moe_grouped_gate_up_kernel and w8a8_block_dynamic_fp8_moe_grouped_down_kernel in which host_descriptor is not implemented.

finegrained-fp8/tests/utils.py
sywangyi: Good point. The CPU path was added to avoid very large temporary fp32 allocations on the accelerator for the biggest MoE shapes, but I agree it regresses CUDA test setup time since CUDA RNG is much faster.

I'll gate this so CUDA keeps generating weights directly on device, and only use the CPU construction path where it is needed to avoid accelerator OOM, e.g. XPU / very large shapes.

sywangyi:
> the config pruner looks great ! not sure about the other changes as the intention is not documented


Besides the GRF config pruner, this PR also updates the W8A8 fused grouped-MoE path for XPU. The descriptor-load path and transposed accumulator layout are intended to improve XPU DPAS utilization for grouped MoE shapes where each expert often receives a small or sparse M tile.

This path is meant to be XPU-specific. CUDA should keep using the existing pointer-compatible load path; the memory-mode gating will drop `device_descriptor` in cuda.


This PR also tries to fix oom for the biggest MoE shapes during test,

IlyasMoutawwakil:
@sywangyi we already have that same idea already implemented for mx paths 😅 can you please try to follow the naming and structure of that implementation, no need to diverge

IlyasMoutawwakil:
also do you have any benchmarks backing the descriptor addition in the block scaled path ? last time i tried i saw no gain from TMA in that path specifically

finegrained-fp8/tests/utils.py
```diff
@@ -87,6 +87,30 @@ def _ue8m0_exp(x):
...
+    # W, padded W and scaled can briefly overlap during device-side quantization.
+    _FP32_WEIGHT_SCRATCH_FACTOR = 3
+    scratch_bytes = numel * torch.empty((), dtype=torch.float32).element_size()
+    scratch_bytes *= _FP32_WEIGHT_SCRATCH_FACTOR
+    return "cpu" if scratch_bytes > free_bytes // 2 else target
+
+
 def make_weights(
     out_features,
     in_features,
```
IlyasMoutawwakil, line 92: can you please compare the make weights latency between with this change vs the full accelerator version, i'd like to avoid slowing down the testing suite as much as possible

sywangyi: I revert the change in test utils, so no impact now.

finegrained-fp8/torch-ext/finegrained_fp8/utils.py
```diff
@@ -254,7 +273,9 @@ def get_accelerator_autotuning_configs(
 
     if tune_block_nk:
         tiles = (
-            [(128, 128)] if is_xpu else [(128, 128), (256, 128), (128, 64), (64, 256)]
+            [(128, 128), (128, 64), (64, 128), (64, 64)]
```
IlyasMoutawwakil, line 280: please add to the docstring like cuda what these block sizes were based on

IlyasMoutawwakil (dismissed):
LGTM ! just one nit

IlyasMoutawwakil (approved):
LGTM ! thanks

PR #1029, opened 2026-07-15
liger-kernels: add device context manager

vasqu:
It's affecting models that use device map auto (under multi gpu setting) and `use_kernels=True`, e.g. RMS Norm gets corrupted. But this goes for any triton kernel launch so it is propogated to all of them

For more details, see https://github.com/linkedin/Liger-Kernel/issues/1303


sayakpaul (dismissed):
I think it applies to all kernels that don't have this fix as you also pointed out in Slack. But I guess we can treat on a case by case basis?

vasqu:
Let me hunt them all down as best as I can, I think it's low effort for big impact tbh so no need to delay on case by case imo

vasqu:
Added it to all triton launches I could find

sayakpaul (approved):

PR #1031, opened 2026-07-16
natten: port NATTEN v0.21.7 to kernel-builder

sayakpaul:
Port https://natten.org/ for an upcoming cool release


natten/build.toml
```diff
@@ -0,0 +1,459 @@
...
+[kernel.natten]
+backend = "cuda"
+cuda-capabilities = [
+    "7.0",
+    "7.5",
+    "8.0",
+    "8.6",
+    "8.9",
+    "9.0",
+    "10.0",
```
alihassanijr, line 26: I noticed SM90 and SM100 have separate builds that compile with their arch-specific features (sm90a and sm100a). What's the reason for this separation?

natten/build.toml
```diff
@@ -0,0 +1,459 @@
...
+    "csrc/autogen/src/cuda/hopper_fmha_bwd/source_2.cu",
+    "csrc/autogen/src/cuda/hopper_fmha_bwd/source_3.cu",
+    "csrc/autogen/src/cuda/hopper_fmha_bwd/source_4.cu",
+]
+
+[kernel.natten_blackwell]
+backend = "cuda"
+cuda-minver = "12.8"
+cuda-capabilities = [
+    "10.0a",
```
alihassanijr, line 366:
```suggestion
cuda-capabilities = [
    "10.0a",
    "10.3a",
```

I'd make this change to support Blackwell Ultra ([G]B300) -- but only if the wheel is being built with CTK 12.9 or later. Sm103 was added in 12.9 release.

NATTEN reference for arches supported based on CTK version: 
https://github.com/SHI-Labs/NATTEN/blob/main/scripts/packaging/pkg_helpers.bash#L43

natten/build.toml
sayakpaul: The sections are how `kernel-builder` scopes `-gencode` flags: `cuda-capabilities` applies per [kernel.*] 
section, so the hopper/blackwell sources get compiled only as `sm_90a` / `sm_100a` while the
legacy+FNA/FMHA core builds for the full 7.0–12.0 range — same effect as the per-source arch flags in
upstream's `setup.py`, just expressed in kernel-builder's schema. 

The blackwell section additionally carries `cuda-minver = "12.8"`, so it's skipped when building the CUDA 12.6 variants (where the SM100 CUTLASS kernels can't compile) without dropping the rest of the kernel. Each section also passes its `NATTEN_WITH_*_FNA` define to match the `#ifdef` gates in the autogen TUs; the core dispatchers derive the same defines from `CUDART_VERSION` so dispatch stays consistent with what was built.

natten/build.toml
sayakpaul: Here's the blocker for that: https://github.com/huggingface/kernels/issues/720

sayakpaul:
@alihassanijr would it be possible for you to help us perform the "building and publishing" step? Our CI is running out: https://github.com/huggingface/kernels-community/actions/runs/29692416939 :D 

Otherwise, I think we could do something like so (to keep moving things forward):

Reduce the core natten section's arch list (`natten/build.toml` currently has 8 entries: 7.0, 7.5, 8.0, 8.6, 8.9, 9.0, 10.0, 12.0 — every one is a `-gencode` pass on each of the 144 autogen TUs). `flash-attn2` in this repo ships just [8.0, 9.0, 10.0, 12.0], relying on within-major SASS forward compatibility (sm_80 code runs on 8.6/8.9). I think dropping 8.6+8.9 alone cuts ~25% of the core section's `nvcc` work; also dropping 7.0/7.5 (no V100/T4) further reduces it. 

Caveat: sm_86/89 have less shared memory than sm_80, so this is safe only if NATTEN's runtime config selection respects the device's actual smem limit.

Not sure if this applies to `natten` but we recently did https://github.com/huggingface/kernels-community/issues/1032 which helped us cut down the build time of FA3 significantly.

alihassanijr:
> @alihassanijr would it be possible for you to help us perform the "building and publishing" step? Our CI is running out: https://github.com/huggingface/kernels-community/actions/runs/29692416939 :D
> 
> Otherwise, I think we could do something like so (to keep moving things forward):
> 
> Reduce the core natten section's arch list (`natten/build.toml` currently has 8 entries: 7.0, 7.5, 8.0, 8.6, 8.9, 9.0, 10.0, 12.0 — every one is a `-gencode` pass on each of the 144 autogen TUs). `flash-attn2` in this repo ships just [8.0, 9.0, 10.0, 12.0], relying on within-major SASS forward compatibility (sm_80 code runs on 8.6/8.9). I think dropping 8.6+8.9 alone cuts ~25% of the core section's `nvcc` work; also dropping 7.0/7.5 (no V100/T4) further reduces it.
> 
> Caveat: sm_86/89 have less shared memory than sm_80, so this is safe only if NATTEN's runtime config selection respects the device's actual smem limit.
> 
> Not sure if this applies to `natten` but we recently did #1032 which helped us cut down the build time of FA3 significantly.

Would love to help @sayakpaul , but as is I'm struggling to find resources to build the standard wheels for 0.21.7 🤣 

But cutting down with +PTX makes sense to me.
And no it shouldn't create any issues, because as far as libnatten is concerned, cutlass-fna/fmha compiles if SM80 and later, hopper-fna/fmha if SM90a, and blackwell-fna/fmha if SM100a/SM103a are specified.

Figuring out shared memory limit, supported tile shapes / configs, and the rest are all determined at runtime based on the actual device.

Beyond that it's just handled by the cuda compiler and how it builds fatbins, with the exception of arch guards we place, but again those are pretty relaxed for cutlass-fna/fmha:

image

image

** Disregard ** the fallthru error message -- it should read "built for SM80 - SM120"

sayakpaul:
> Would love to help @sayakpaul , but as is I'm struggling to find resources to build the standard wheels for 0.21.7 🤣

Is it related to `kernels`? To me, it seems like related to how you distribute the wheels?

Noted regarding the rest of what you said. Shall I wait for you till you build?

alihassanijr:
> > Would love to help @sayakpaul , but as is I'm struggling to find resources to build the standard wheels for 0.21.7 🤣
> 
> Is it related to `kernels`? To me, it seems like related to how you distribute the wheels?
> 
> Noted regarding the rest of what you said. Shall I wait for you till you build?

Nope, unrelated to kernels. I just don't have a standard way of building wheels. Every release I try to find resources to build, sometimes it's easier and quicker, sometimes it's more involved.

For this release, I'm just going to run the full pipeline to make sure it's stable, and tag so that `kernels` can pick up the latest fixes and perf improvements, and then figure out where/how to build wheels.

sayakpaul:
Sounds good! Keep me posted.

alihassanijr:
@sayakpaul 0.21.7 is tagged and live, wheels are also up:

https://github.com/SHI-Labs/NATTEN/releases/tag/v0.21.7

sayakpaul:
Also, ran a benchmark on H100 and here's the report: https://gist.github.com/sayakpaul/8031209813d71fd19ef4882f67c08df6

danieldk (approved):

PR #1033, opened 2026-07-17
feat: report kernel bot upload results

drbh:
this PR implements two release upload features that build on recent `kernel-builder` changes.

first, it uses the `--create-pr` flag added in https://github.com/huggingface/kernels/pull/694 when a kernel uploads to a Hub repo outside `kernels-community`. this lets CI upload kernels like `MiniMaxAI/msa` and `sgl-project/sgl-flash-attn3` by opening Hub PRs instead of assuming we can push directly.

second, it uses the upload json output added in https://github.com/huggingface/kernels/pull/716 so each build can write its upload result, the workflow can aggregate the results after the matrix completes, and the kernel bot can update its existing comment with the Hub repo / Hub PR links.

the main plumbing is:

- resolve upload args from `build.toml`, using `--create-pr` for external repos and `--repo-id` for community repos
- write `kernel-builder upload --output-json` results as short lived workflow artifacts
- pass the kernel bot comment id through dispatch and update that comment from `report-status`
- use the external Hub repo ids for `msa` and `sgl-flash-attn3`

### added script usage

this script checks if the build should be a PR or not and creates the args needed to be passed to the upload command
```bash
uv run .github/scripts/hub_pr_upload_args.py upload-args msa kernels-community
```

outputs
```
--create-pr
````

and this script reads the json output and structures it into text that is appended to the kernel bot's last comment
```bash
uv run .github/scripts/format_hub_upload_comment.py msa kernels-community '{"pull_requests":[{"url":"https://hf.co/kernels/MiniMaxAI/msa/discussions/1"}]}'
```

outputs
```
Hub uploads:
- Hub repo: https://huggingface.co/kernels/MiniMaxAI/msa
- Hub pull request: https://hf.co/kernels/MiniMaxAI/msa/discussions/1
```

also ran the dispatch/pr bot tests, yaml parsing, py_compile, and diff whitespace checks.

TLDR; this PR improves uploads to respect creating PR's and better reporting on the results

### DIFF
--- a/.github/scripts/format_hub_upload_comment.py
+++ b/.github/scripts/format_hub_upload_comment.py
@@ -1,0 +1,77 @@
+#!/usr/bin/env python3
+# Input:
+#   [--base <comment-body>] <kernel> <repo-prefix> <upload-json-or-path>...
+#
+# Output:
+#   A Hub upload section, or <comment-body> with that section updated.
+#
+# Example:
+#   python3 .github/scripts/format_hub_upload_comment.py msa kernels-community "$RUNNER_TEMP"/upload*.json
+#   python3 .github/scripts/format_hub_upload_comment.py --base "$BODY" msa kernels-community "$RUNNER_TEMP"/upload*.json
+#   python3 .github/scripts/format_hub_upload_comment.py msa kernels-community '{"pull_requests":[{"url":"https://hf.co/kernels/MiniMaxAI/msa/discussions/1"}]}'
+import argparse
+import json
+from pathlib import Path
+
+from hub_pr_upload_args import repo_id
+
+SECTION = "Hub uploads:"
+
+
+def load_json(value):
+    if not value.lstrip().startswith(("{", "[")):
+        path = Path(value)
+        if path.is_file():
+            with open(path) as f:
+                return json.load(f)
+    return json.loads(value)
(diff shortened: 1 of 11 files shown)


.github/scripts/format_hub_upload_comment.py
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env python3
+# Input:
+#   [--base <comment-body>] <kernel> <repo-prefix> <upload-json-or-path>...
+#
+# Output:
+#   A Hub upload section, or <comment-body> with that section updated.
```
sayakpaul, line 6: (nit): Possible to include an actual representative example?

sayakpaul (approved):
Very cool! 

@vasqu FYI. I think we should also do the same for https://github.com/huggingface/kernels-community/pull/1010

.github/scripts/format_hub_upload_comment.py
drbh: just below this section are some copy pastable commands to test the script and see the output. for example in the PR description above there is an example running the script and its output

I avoided including the string output in the comment in the case we tweak the formatting - however it should be very easy to run the command to see what the output would look like using on of the examples in the header comments

PR #1016, opened 2026-07-07
feat: filter kernel bot build backends

drbh:
the PR updates the kernel bot to enable filtering the backends to target

```bash
python .github/scripts/pr_comment_kernel_bot.py \
--dry-run "/kernel-bot build flash-attn2[xpu,cpu]" \
--pr-number 1007
```
outputs
```
Skipping Windows build for flash-attn2 (not in WINDOWS_KERNELS allowlist)

[dry-run] build.yaml:
{
  "ref": "main",
  "inputs": {
    "kernel_name": "flash-attn2",
    "dispatch_key": "pr1007-flash-attn2-build.yaml-3b9a156aa092",
    "mode": "release",
    "backends": "cpu,xpu",
    "repo_prefix": "kernels-community",
    "pr_number": "1007",
    "head_sha": "f80fa0a5bdcae73fe8549f57abed832eb6e304e6",
    "target_branch": "pr-1007",
    "upload": "false"
  }
}
```

and with no filter

```bash
python .github/scripts/pr_comment_kernel_bot.py \
--dry-run "/kernel-bot build flash-attn2" \
--pr-number 1007
```
outputs
```
Skipping Windows build for flash-attn2 (not in WINDOWS_KERNELS allowlist)

[dry-run] build.yaml:
{
  "ref": "main",
  "inputs": {
    "kernel_name": "flash-attn2",
    "dispatch_key": "pr1007-flash-attn2-build.yaml-bd3fc94efa15",
    "mode": "release",
    "backends": "cpu,cuda,xpu",
    "repo_prefix": "kernels-community",
    "pr_number": "1007",
    "head_sha": "f80fa0a5bdcae73fe8549f57abed832eb6e304e6",
    "target_branch": "pr-1007",
    "upload": "false"
  }
}
```

PR #1036, opened 2026-07-21
feat: prefer reading the build toml from current ref

drbh:
Re-targets #982 at `main`. That PR was merged into `kernel-bot-refactor`, which had already been squash-merged and closed via #983 three weeks earlier — so the change never reached `main`.

The kernel bot checks out the default branch, so it read `build.toml` from `main` rather than from the PR. A kernel whose `build.toml` is added or whose `backends` change in the PR got the wrong workflow set — a new cuda-only kernel would still dispatch `build-mac.yaml`.

`read_backends` now takes an optional `ref` and reads via `git show <ref>:<kernel>/build.toml`. Empty ref keeps the working-tree read, so dry-run and local behavior are unchanged. `pr-comment-build.yaml` fetches the PR head at depth 1 so the blob is available.

Cherry-picked onto current `main`, which required reconciling with #1016: `select_workflows` there takes `backends` as a parameter so `parse_kernel_arg` can support inline overrides (`relu[cuda,rocm]`). Kept that design and threaded `metadata_ref` only through `read_backends` — applying the original version as-written would have bypassed the requested-backends filter. Tests rewritten from `unittest` to pytest to match `main`.

57 tests pass.

### DIFF
--- a/.github/scripts/dispatch.py
+++ b/.github/scripts/dispatch.py
@@ -109,12 +109,25 @@ def parse_kernel_arg(token: str) -> tuple[str | None, list[str] | None]:
     return name, backends
 
 
-def read_backends(kernel_name: str) -> list[str] | None:
-    build_toml = Path(kernel_name) / "build.toml"
-    if not build_toml.exists():
-        return None
-    with open(build_toml, "rb") as f:
-        config = tomllib.load(f)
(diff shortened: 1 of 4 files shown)


drbh:
re merging as the original PR was approved and this branch just corrects where the changes are merged

PR #1040, opened 2026-07-28
Enable xpu kernels on CRI

jiqing-feng:
In progress

### DIFF
--- a/flash-attn2/flash_attn_xpu/flash_api.cpp
+++ b/flash-attn2/flash_attn_xpu/flash_api.cpp
@@ -120,6 +120,10 @@ mha_fwd(
     at::Tensor out_padded;
     if (out_.has_value()) {
         auto out_val = out_.value();
+        // The kernel writes out_dtype elements; a mismatched buffer would be
+        // overrun (MXFP writes float32 into what may be a half-width tensor).
+        TORCH_CHECK(out_val.scalar_type() == out_dtype,
+                    "out must have dtype ", out_dtype, ", got ", out_val.scalar_type());
         if (needs_padding) {
             const int pad_size = head_size_padded - head_size_actual;
             out_padded = torch::nn::functional::pad(out_val, torch::nn::functional::PadFuncOptions({0, pad_size}));
@@ -496,6 +500,10 @@ mha_varlen_fwd(
     auto out_dtype = is_mxfp ? torch::kFloat32 : c10::typeMetaToScalarType(q_dtype);
     at::Tensor out_padded;
     if (out_.has_value()) {
+        // The kernel writes out_dtype elements; a mismatched buffer would be
+        // overrun (MXFP writes float32 into what may be a half-width tensor).
+        TORCH_CHECK(out_.value().scalar_type() == out_dtype,
+                    "out must have dtype ", out_dtype, ", got ", out_.value().scalar_type());
         out_padded = maybe_pad(out_.value());
     } else if (is_mxfp) {
         out_padded = torch::zeros({total_q, num_heads, head_size_padded},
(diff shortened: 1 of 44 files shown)


flash-attn2/torch-ext/torch_binding.cpp
```diff
@@ -36,7 +36,10 @@ TORCH_LIBRARY_EXPAND(TORCH_EXTENSION_NAME, ops) {
     "int window_size_right, "
     "float softcap, "
     "bool return_softmax, "
-    "Generator? gen_) -> Tensor[]");
+    "Generator? gen_, "
+    "Tensor? scale_q_=None, "
```
YangKai0616, line 40: The public custom ops and `flash_attn_func/flash_attn_varlen_func` neither accept nor forward them, should the Python API be updated synchronously?

flash-attn2/flash_attn_xpu/src/fmha_fwd.hpp
```diff
@@ -66,9 +90,21 @@ inline void policy_dispatch(
     const fmha_fwd_args_t& args) {
   if (cuType == CutlassType::half) {
     policy_dispatch_fp16<chunk_policy, PipelineStages, IsVarLen, IsPaged, HasRotary>(queue, args);
-  } else {
-    policy_dispatch_bf16<chunk_policy, PipelineStages, IsVarLen, IsPaged, HasRotary>(queue, args);
+    return;
   }
+  if constexpr (!IsPaged && !HasRotary) {
```
YangKai0616, line 95: Paged or rotary MXFP skips this branch and then falls through to `policy_dispatch_bf16`. We could reject unsupported MXFP combinations before dispatch.

flash-attn2/flash_attn_xpu/flash_api.cpp
```diff
@@ -65,44 +68,67 @@ mha_fwd(
...
     if (needs_padding) {
-        const int pad_size = head_size_padded - head_size_og;
+        const int pad_size = head_size_padded - head_size_actual;
         q_padded = torch::nn::functional::pad(q, torch::nn::functional::PadFuncOptions({0, pad_size}));
         k_padded = torch::nn::functional::pad(k, torch::nn::functional::PadFuncOptions({0, pad_size}));
         v_padded = torch::nn::functional::pad(v, torch::nn::functional::PadFuncOptions({0, pad_size}));
     }
 
+    // Low-precision (FP8/FP4) inputs accumulate in float32 and produce a float32 output.
+    auto out_dtype = is_mxfp ? torch::kFloat32 : c10::typeMetaToScalarType(q_dtype);
```
YangKai0616, line 118: The real MXFP path now returns FP32, with twice the packed Q width for FP4, but both fake implementations still use `torch.empty_like(q)`, therefore their dtype and FP4 shape disagree with runtime. There will be issues when using `torch.compile`, right?

flash-attn2/flash_attn_xpu/src/fmha_fwd.hpp
jiqing-feng: Agreed, and the fall-through is the worst kind — it reinterprets the fp8/fp4 bytes as bf16 and returns wrong numbers without any error. Of the two combinations, **rotary is already safe**: `mha_fwd_kvcache` (the only `HasRotary=true` entry point) still restricts `q_dtype` to fp16/bf16, so MXFP cannot reach it. **Paged was the real hole** — `mha_varlen_fwd` accepts MXFP `q` together with a `block_table`. Added an explicit guard next to `is_paged`.
Kept the check at the API boundary rather than inside `policy_dispatch`, so the `if constexpr` guard can keep doing its real job (avoiding instantiation of the expensive MXFP kernels on the paged/rotary template paths).

flash-attn2/flash_attn_xpu/flash_api.cpp
jiqing-feng: Correct on both counts, and yes — `torch.compile` would propagate the fp8/fp4 dtype (and, for fp4, the packed width) downstream and then hit a mismatch against the real kernel output. Fixed in both `_flash_attn_forward_fake` and `_flash_attn_varlen_forward_fake`
This mirrors the host side: `head_size_actual = fp4 ? head_size_og * 2 : head_size_og`, MXFP always allocates with `out_dtype = kFloat32`, and fp4 is explicitly excluded from padding (`needs_padding && q_dtype != kFloat4_e2m1fn_x2`), so the returned width is exactly `head_size_actual`.

Also related, in the same area: `mha_fwd` / `mha_varlen_fwd` now `TORCH_CHECK` that a user-supplied `out_` has dtype `out_dtype`. Previously an fp16 `out_` would be overrun by the fp32 kernel writes.

PR #1046, opened 2026-07-29
esmfold2-trimul: add ESMFold2 triangle multiplication kernel

Rocketknight1:
[deleting like 500 lines of claudeslop] This is the PR for the esmfold2 trimul kernel! We already have a triangular multiplication kernel in the repo but it's quite different from esmfold2's in terms of outputs, so I wasn't able to use it in the PR. It's also critical; performance is really badly affected without it enabled.


sayakpaul:
/kernel-bot build-and-stage esmfold2-trimul-kernel --branch v0

esmfold2-trimul-kernel/build.toml
```diff
@@ -0,0 +1,12 @@
+[general]
+name = "esmfold2-trimul-kernel"
```
sayakpaul:
```suggestion
name = "esmfold2-trimul"
```

WDYT?

esmfold2-trimul-kernel/build.toml
```diff
@@ -0,0 +1,12 @@
...
+version = 1
+license = "Apache-2.0"
+backends = ["cuda"]
+
+[general.hub]
+repo-id = "kernels-community/esmfold2-trimul-kernel"
```
sayakpaul: Did we author this kernel? If not, it could be nice to use the actual org that did (if we have a collaboration with them). Our CI can then discover that the kernel isn't from `kernels-community` and can _automatically_ open a PR to the specified kernels repo on the Hub.

sayakpaul (commented):
Left some comments.

sayakpaul:
Check if this kernel works https://huggingface.co/kernels/kernels-staging/esmfold2-trimul-kernel (you will have to provide the appropriate repo id of course).

esmfold2-trimul-kernel/build.toml
Rocketknight1: We didn't, but the original source isn't on the Hub! It's from BioHub, they included it in their fork of Transformers for developing ESMFold2/ESMC. So the only kernel Hub repos are uploaded by me.

sayakpaul: Then we should probably have the kernel hosted under ESMFold2 org and specify the `repo_id` to be `ESMFold2/esmfold2-trimul-kernel`

Rocketknight1:
@sayakpaul Accepted your suggestion, and the new repo works!

sayakpaul:
Seems pending https://github.com/huggingface/kernels-community/pull/1046#discussion_r3682438090 ?

Rocketknight1:
@sayakpaul Makes sense! Wouldn't we need to mark them as a trusted org for kernels, though? Or how does that work?

sayakpaul:
So, we would have to first ship the kernel to `kernels-community`, then ask the org collaborators to check if they're okay with the transfer. Once approved, then we can internally grant them the "Trusted Publisher" status and transfer the kernel to their org.

sayakpaul (dismissed):

sayakpaul:
/kernel-bot build-and-stage esmfold2-trimul-kernel --branch v0

sayakpaul:
@Rocketknight1 the builds are failing: https://github.com/huggingface/kernels-community/actions/runs/30559678877

Rocketknight1:
@sayakpaul Renamed everything to `esmfold2-trimul`, there were a few string references still using `esmfold2-trimul-kernel`

esmfold2-trimul/tests/test_esmfold2_trimul.py
```diff
@@ -0,0 +1,182 @@
+import pytest
+import torch
+import torch.nn.functional as F
+
+import esmfold2_trimul
```
sayakpaul, line 5: We should be using `get_kernel()` here, though. Do we do that here or batch update it as a part of https://github.com/huggingface/kernels-community/issues/1050. Question for @danieldk

sayakpaul (approved):

PR #1047, opened 2026-07-29
fla: add gdn kernels

vasqu:
As per title

### DIFF
--- a/fla/build.toml
+++ b/fla/build.toml
@@ -1,6 +1,6 @@
 [general]
 name = "fla"
-version = 1
+version = 2
 license = "MIT"
 backends = [
   "cuda",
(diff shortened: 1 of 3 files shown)


fla/build.toml
```diff
@@ -1,6 +1,6 @@
 [general]
 name = "fla"
-version = 1
+version = 2
```
sayakpaul: But we aren't updating signatures no?

sayakpaul (commented):
Just one comment. Additionally, would it be possible to change the `repo_id` in `build.toml` so that the kernel points to the actual org now that we can open PRs? Or not yet?

fla/build.toml
vasqu: Force of habit will revert :grimacing:

vasqu:
I don't think they have responded yet so we don't have a respective hub repo on their side :cry:

danieldk (approved):

PR #1051, opened 2026-07-30
mamba-ssm: add layers

vasqu:
As per title WIP, adding layer variations for mamba2 and mamba1 so we can deprecate lazy load kernels

- [x] Mamba2
- [x] Mamba1

Also fixes a multi gpu issue under triton (the same we had under liger), ref. https://github.com/state-spaces/mamba/pull/1003

### DIFF
--- a/mamba-ssm/torch-ext/mamba_ssm/layers.py
+++ b/mamba-ssm/torch-ext/mamba_ssm/layers.py
@@ -107,6 +107,7 @@ class selective_state_update(nn.Module):
         D: torch.Tensor | None = None,
         dt_bias: torch.Tensor | None = None,
         dt_softplus: bool = False,
+        z: torch.Tensor | None = None,
         **kwargs,
     ):
         return cuda_selective_state_update(
@@ -117,7 +118,7 @@ class selective_state_update(nn.Module):
             B,
             C,
             D,
-            z=None,
+            z=z,
             dt_bias=dt_bias,
             dt_softplus=dt_softplus,
         )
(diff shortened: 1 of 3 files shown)


mamba-ssm/torch-ext/mamba_ssm/ops/triton/ssd_chunk_scan.py
vasqu, line 1: Triton corruption may happen otherwise; similar to liger a few days/weeks back

A lot of ppl did different fixes but never the true underlying fix

sayakpaul (dismissed):

mamba-ssm/torch-ext/mamba_ssm/layers.py
```diff
@@ -43,4 +240,12 @@ def forward(
...
-__all__ = ["causal_conv1d_fn", "causal_conv1d_update"]
+__all__ = [
+    "causal_conv1d_fn",
+    "causal_conv1d_update",
+    "mamba_inner_fn",
+    "mamba_split_conv1d_scan_combined",
+    "mamba_chunk_scan_combined",
+    "selective_state_update",
+    "selective_scan_fn",
+]
```
danieldk, line 251: Adding functions on a compiled kernels (so not no-arch like `torch-cuda`) requires a version bump, since otherwise, there will be old variants (Torch 2.10, Torch 2.11) that do not have these functions.

danieldk (approved):
Looks awesome!

PR #1057, opened 2026-08-03
aiter-flash-attn: support sliding-window attention (fixes #1055)

### LINKED ISSUE #1055: aiter-flash-attn: re-vendor newer aiter to enable sliding-window attention (ROCm/aiter#3742)
## Summary

`aiter-flash-attn` currently rejects any call with `window_size_right != -1`, so sliding-window
attention models (Gemma3, Mistral, Qwen2 SWA variants, ...) cannot use this kernel through
`transformers`' `FLASH_ATTN_KERNEL_FALLBACK` ROCm entry. They fail with:

```
ValueError: window_size_right is not supported yet in the Triton Backend
```

Upstream aiter fixed this on 2026-06-29, but the vendored copy here predates it.

## Version provenance

The published Hub artifact `kernels-community/aiter-flash-attn` reports
`"id": "_aiter_flash_attn_rocm_e5d81f8"` in `build/torch-rocm/metadata.json`, which corresponds to
`e5d81f8da` ("aiter-flash-attn: ship MHA configs in Hub artifact", #912, 2026-06-08).

The relevant upstream change is ROCm/aiter commit
[`13040146d`](https://github.com/ROCm/aiter/commit/13040146d) ("[triton] Add sliding-window support
to flash-attention backward", ROCm/aiter#3742, 2026-06-29) — three weeks after the vendored
snapshot.

Separately, the Hub artifact also appears to be behind this repo's own source: `aiter-flash-attn/`
has commits from 2026-07-04, 07-06 and 07-08 (RDNA3.5 configs, Torch 2.13 fixes) that are not
reflected in the published build. A republish looks due regardless of this issue.

## Current behaviour

`torch-ext/aiter_flash_attn/mha.py` guards unconditionally, and additionally blocks the `dao_ai`
path that would otherwise handle it:

```python
if window_size_right != -1:
    raise ValueError("window_size_right is not supported yet in the Triton Backend")
...
assert (
    window_size_left == -1 and window_size_right == -1
), "dao_ai impl does not support sliding window attention."
```

## Upstream fix

ROCm/aiter#3742 makes the guard conditional, drops the `dao_ai` assert, and stops hardcoding
`window_size_right=-1` when forwarding into `flash_attn_2` (forward and backward):

```diff
-    if window_size_right != -1:
+    if _MHA_IMPL != "dao_ai" and window_size_right != -1:
         raise ValueError("window_size_right is not supported yet in the Triton Backend")

-            window_size_left == -1 and window_size_right == -1
-        ), "dao_ai impl does not support sliding window attention."

-                window_size_right=-1,
+                window_size_right=window_size_right,
```

Note the default impl still raises; SWA is supported only on the `dao_ai`
(`flash_attn_triton_amd`) path, which is already vendored here.

I verified the updated upstream code on AMD ROCm hardware (dense, varlen and backward, matching an SDPA
sliding-window reference within bf16 tolerance), so the upstream fix does work in practice.

## Request 1: re-vendor at a newer aiter revision

Syncing `torch-ext/aiter_flash_attn/` to current aiter `main` is roughly 1,379 changed lines across
13 files (imports normalised), concentrated in the `dao_ai` backend:

| file | changed lines |
| --- | --- |
| `_kernels/flash_attn_triton_amd/bwd.py` | 458 |
| `_kernels/flash_attn_triton_amd/fwd_prefill.py` | 165 |
| `_kernels/flash_attn_triton_amd/fwd_decode.py` | 133 |
| `_kernels/flash_attn_triton_amd/interface_v3.py` | 115 |
| `_kernels/flash_attn_triton_amd/interface_v2.py` | 101 |
| `mha.py` | 98 |
| 7 other files | 309 |

Since `build.toml` pins no aiter revision, I could not tell whether the vendoring is scripted. Two
options, and I'd rather follow your preference than guess:

1. **Full re-vendor** at a named aiter revision. Faithful to the "self-contained repackaging" goal,
   but pulls in a month of unrelated `dao_ai` churn.
2. **Surgical port** of ROCm/aiter#3742 into `mha.py` only (~10 lines). Minimal risk, but introduces
   deliberate drift from upstream.

Either way, I can only validate on one of the supported architectures; covering the full set listed
in the kernel README would need maintainers or others with that hardware.

## Request 2: auto-select `dao_ai` when a right window is requested

Re-vendoring alone does not change observable behaviour, because nothing calls `mha_set_impl`.
`transformers` imports only `flash_attn_func` / `flash_attn_varlen_func`, so the default impl — which
still raises — is what runs.

`dao_ai` also cannot simply become the global ROCm default: it asserts out on attention sinks,
positional encoding (`pe_head_dim > 0`) and FP8, so sink models such as gpt-oss would break.

A natural fix is to dispatch per call inside the vendored wrapper rather than pushing impl selection
up into `transformers`:

```python
impl = _MHA_IMPL
if impl == "default" and window_size_right != -1 and sink is None and pe_head_dim == 0 and not IS_FP8:
    impl = "dao_ai"
```

This keeps sink/FP8 workloads on the default kernel, needs no `transformers` change, and leaves
`mha_set_impl` available as an explicit override. It is a deviation from upstream aiter semantics,
though, so it's your call whether it belongs here or should be proposed to ROCm/aiter instead.

Abdennacer-Badaoui:
## What

Enables sliding-window attention on ROCm so SWA models ( Gemma3, Mistral, Qwen2 SWA, and
**gpt-oss** (SWA + attention sinks) ) work through the plain `flash_attn_func` / `flash_attn_varlen_func` paths `transformers` uses. Previously `window_size_right != -1` raised `ValueError: window_size_right is not supported yet`. fixes #1055.

## How

- **Re-vendored** `torch-ext/aiter_flash_attn/` from `ROCm/aiter@8a9186d`, bringing real SWA support incl. the sliding-window **backward** (aiter #3742) the old snapshot predated.
- **Per-call routing** in `mha.py` (forward + both backwards): under `causal` a right window is a no-op, so it's normalized to `-1` and stays on the default kernel (which supports left window **+ sinks**, so gpt-oss works); a *non-causal* right window auto-selects the `dao_ai` backend when compatible (no sink/PE/FP8).

## Testing

Real AMD **gfx942** (MI300-class): 12/12 pass; SWA forward, causal, varlen, **backward** vs SDPA, and gpt-oss-style causal SWA **+ sinks**. Numerics verified on gfx942 only.

### DIFF
--- a/aiter-flash-attn/README.md
+++ b/aiter-flash-attn/README.md
@@ -61,10 +61,29 @@ out = flash_attn.flash_attn_func(q, k, v, causal=True)
 ## Origin
 
 Code is taken from `aiter/ops/triton/attention/mha.py` and its transitive
-imports, with the `dao_ai` impl path stripped (it depended on a separate
-`flash_attn_triton_amd` subpackage we don't need here). All `from aiter.*`
-absolute imports have been rewritten to package-relative form per the
-[Hub kernel requirements](https://huggingface.co/docs/kernels/kernel-requirements).
+imports (including the `flash_attn_triton_amd` / `dao_ai` backend), vendored at
+[`ROCm/aiter@8a9186d`](https://github.com/ROCm/aiter/commit/8a9186d983ed34ece2f68fe039e07f1b0abe147b).
+All `from aiter.*` absolute imports have been rewritten to package-relative form
+per the [Hub kernel requirements](https://huggingface.co/docs/kernels/kernel-requirements).
+
+Two intentional local deviations from upstream:
+
+- `utils/_triton/arch_info.py` detects the GPU arch lazily (on first `get_arch()`
+  call) so the module imports in the kernel-builder Nix sandbox, which has no
+  active GPU driver.
+- `mha.py` resolves sliding-window attention per call so it works through the
+  plain `flash_attn_func` / `flash_attn_varlen_func` entry points without an
+  explicit `mha_set_impl`:
+    - Under `causal=True` a right window (`window_size[1] >= 0`) is a no-op — the
+      causal edge already bounds the right side — so it is normalized to `-1` and
+      the call stays on the default kernel, which supports a left window together
+      with attention sinks. This keeps causal sliding-window models on the tuned
+      default path, **including sink models such as gpt-oss** (Gemma3, Mistral,
+      Qwen2 SWA also land here).
+    - A non-causal right window (which the default kernel cannot express) auto-
+      selects the `dao_ai` backend, but only when the workload is compatible with
+      it (no attention sink, no positional-encoding head split, not FP8).
+      Sink/PE/FP8 workloads stay on the default kernel.
 
 ## License
 
--- a/aiter-flash-attn/flake.lock
+++ b/aiter-flash-attn/flake.lock
@@ -41,11 +41,11 @@
         "rust-overlay": "rust-overlay"
       },
       "locked": {
-        "lastModified": 1783417137,
-        "narHash": "sha256-/iWe/m5ZACdb/DcXTrvXmbHS2QrzgGcGa7hPKi460tE=",
+        "lastModified": 1785743322,
+        "narHash": "sha256-6iiCDJZ1KJd+QdS9LOln2ZleEyFMK94/Q24Gsbb0bEI=",
         "owner": "huggingface",
         "repo": "kernels",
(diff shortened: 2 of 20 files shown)


danieldk:
Nice! I tried to figure out whether this requires a version bump. I think it doesn't, it does add functionality, but there are no other builds lingering around that wouldn't get the change since it's a no-arch kernel.

danieldk (approved):

PR #1059, opened 2026-08-04
megablocks MXFP8 and Crescent Island support

jiqing-feng:
Adds MXFP8 to the megablocks XPU grouped GEMM and makes both megablocks and
flash-attn2 build and run on Intel Crescent Island (CRI, Xe35).

Depends on the build-side changes in `kernels#734`.

## MXFP8

Block-scaled FP8 (`e4m3`/`e5m2` weights with `e8m0` scales, group size 32/64/128)
now goes through the same block-scaled kernel path as the existing MXFP4 support,
so it costs a dtype branch and a scale-offset rule rather than a new kernel.
Weights are `[E, N, K]` and unpacked, unlike the 4-bit path.

This adds an `is_B_mxfp8` argument to `cutlass_grouped_gemm_interface`, so the op
schema changes.

## Why the GEMM is built twice

sycl-tla only emits the Xe35 block-scaled DPAS instructions when
`__SYCL_TARGET_INTEL_GPU_CRI__` is defined, and the compiler never defines it on
its own. Compiling with it on produces code that `ocloc` rejects for pvc and bmg,
so one translation unit cannot cover both generations.

The grouped GEMM is therefore compiled twice from the same sources — an Xe20 unit
for pvc/bmg and a CRI-only unit with the macro forced on and `spir64_gen` as the
sole target — and the host picks between them by the device IP version reported by
the driver. Both units cover the same dtypes, so nothing is lost on older GPUs.

Two things follow from compiling the same headers twice:

- the shared headers live in an inline namespace named after the target
  (`xe20` / `xe35`), otherwise the linker merges symbols that share a name but not
  a body;
- the SYCL kernel name tag carries the arch, otherwise the two device images
  collide.

A `static_assert` ties the instantiation to `SYCL_INTEL_TARGET`, so a build that
loses the macro fails instead of silently producing an Xe20 kernel labelled Xe35.

## flash-attn2

Only a build fix: one reduction in the forward epilogue needed explicit tensor
access to compile against sycl-tla 0.9.2. The MXFP forward path that earlier
commits in this branch added has been reverted, so flash-attn2 ships unchanged
functionally.

## Validation

Built both kernels against `kernels#734` (`torch213-cxx11-xpu20260-x86_64-linux`)
and checked numerics on BMG hardware and the CRI TBX simulator. Each case runs
only the kernel on XPU; data generation, the fp32 reference, dequantization and
error statistics stay on CPU.

- megablocks grouped GEMM — 5/5 pass (bf16 ±bias, fp16, mxfp4, mxfp8), max
  relative error 2.6e-03.
- flash-attn2 — 8/8 pass (fix / varlen × fp16 / bf16 × causal / non-causal), max
  abs error 5.4e-04 (fp16) / 4.0e-03 (bf16).


sayakpaul:
Could you separate the changes for the two kernels: FA2 and MegaBlocks?

jiqing-feng:
> Could you separate the changes for the two kernels: FA2 and MegaBlocks?

Sure.

jiqing-feng:
Separated to #1065 and #1066

PR #1063, opened 2026-08-05
Add Marc to verified contributors

SunMarc:
## What does this PR do ? 

As per title @danieldk . I will soon open a PR to add gguf kernels in kernels-community so that we can run gguf models efficiently in transformers.


SunMarc:
cc @sayakpaul

sayakpaul (approved):

PR #1065, opened 2026-08-06
flash-attn2: fix forward epilogue reduction for sycl-tla 0.9.2

jiqing-feng:
Depends on the build-side changes in
[huggingface/kernels#734](https://github.com/huggingface/kernels/pull/734).

Only a build fix: one reduction in the forward epilogue needed explicit tensor
access to compile against sycl-tla 0.9.2.

## Validation

Built (`torch213-cxx11-xpu20260-x86_64-linux`) and checked numerics on BMG
hardware and the CRI TBX simulator. Each case runs only the kernel on XPU; data
generation, the fp32 reference and error statistics stay on CPU.

- flash-attn2 — 8/8 pass (fix / varlen × fp16 / bf16 × causal / non-causal), max
  abs error 5.4e-04 (fp16) / 4.0e-03 (bf16).


jiqing-feng:
Hi @danieldk . This PR can be reviewed now. Thanks!

danieldk (dismissed):

danieldk (approved):

PR #1066, opened 2026-08-06
Megablocks mxfp8 cri

jiqing-feng:
Depends on the build-side changes in
[huggingface/kernels#734](https://github.com/huggingface/kernels/pull/734).

## MXFP8

Block-scaled FP8 (`e4m3`/`e5m2` weights with `e8m0` scales, group size 32/64/128)
now goes through the same block-scaled kernel path as the existing MXFP4 support,
so it costs a dtype branch and a scale-offset rule rather than a new kernel.
Weights are `[E, N, K]` and unpacked, unlike the 4-bit path.

This adds an `is_B_mxfp8` argument to `cutlass_grouped_gemm_interface`, so the op
schema changes.

## Why the GEMM is built twice

sycl-tla selects its Xe35 code paths on `__SYCL_TARGET_INTEL_GPU_CRI__`, and the
compiler never defines it on its own. The macro reaches through the headers, so
a unit built with it is an Xe35 unit throughout and cannot also cover pvc and
bmg.

The grouped GEMM is therefore compiled twice from the same sources — an Xe20 unit
for pvc/bmg and a CRI-only unit with the macro forced on and `spir64_gen` as the
sole target — and the host picks between them by the device IP version reported by
the driver. Both units cover the same dtypes, so nothing is lost on older GPUs.

Two things follow from compiling the same headers twice:

- the shared headers live in an inline namespace named after the target
  (`xe20` / `xe35`), otherwise the linker merges symbols that share a name but not
  a body;
- the SYCL kernel name tag carries the arch, otherwise the two device images
  collide.

A `static_assert` ties the instantiation to `SYCL_INTEL_TARGET`, so a build that
loses the macro fails instead of silently producing an Xe20 kernel labelled Xe35.

## Validation

Built against `kernels#734` (`torch213-cxx11-xpu20260-x86_64-linux`) and checked
numerics on BMG hardware and the CRI TBX simulator. Each case runs only the
kernel on XPU; data generation, the fp32 reference, dequantization and error
statistics stay on CPU.

- megablocks grouped GEMM — 5/5 pass (bf16 ±bias, fp16, mxfp4, mxfp8), max
  relative error 2.6e-03.

The dispatch was checked rather than assumed: on the simulator the driver reports
IP version 35.11.0, the runtime loads the CRI AOT image instead of falling back
to JIT, and the kernel that launches is `MoE::xe35::GemmCuteName<..., 35>`.

One thing worth stating: MXFP8 here is a software dequantization feeding a plain
`dpas`. The Xe35 block-scaled `bdpas` instructions are not used, so that remains
available as a later optimization.


megablocks/torch-ext/torch_binding.cpp
```diff
@@ -263,7 +263,7 @@ TORCH_LIBRARY_EXPAND(TORCH_EXTENSION_NAME, ops) {
       "cutlass_grouped_gemm_interface(Tensor ptr_A, Tensor ptr_B, "
       "Tensor? ptr_scales, Tensor? ptr_bias, Tensor! ptr_D, "
       "Tensor expert_first_token_offset, int N, int K, int num_experts, "
-      "bool is_B_int4, bool is_B_mxfp4) -> Tensor");
+      "bool is_B_int4, bool is_B_mxfp4, bool is_B_mxfp8) -> Tensor");
```
danieldk: This might warrant a version bump in `build.toml` Not 100% sure though, do we expect that people use this op directly?

jiqing-feng: Good question. The op isn't meant to be used directly — inside the package it is only
reached through `xpu_fused_moe.py` (`cutlass_grouped_gemm` / `cutlass_grouped_gemm_xe2`,
plus the `register_fake` shim), and nothing from that module is re-exported in
`megablocks/__init__.py`. Consumers go through the `MegaBlocksMoeMLP` layer, whose
signature is unchanged.

To be safe for anyone calling the op directly anyway, I gave the new argument a default:

```
bool is_B_int4, bool is_B_mxfp4, bool is_B_mxfp8=False
```

The schema change is now purely additive — the previous 11-argument call still resolves,
so no caller breaks and a `v2` branch isn't needed. A major bump would force every
consumer to move `get_kernel(..., version=1)` to `2` for an argument they never pass.

danieldk (dismissed):

danieldk (approved):

PR #1071, opened 2026-08-10
activation: Add XPU backend to `kernels-community/activation`

jiqing-feng:
## Summary

Adds a SYCL implementation of the `activation` kernels so the component can be
used on Intel GPUs. All eleven operators exposed by the CUDA backend are
implemented and registered under the `XPU` dispatch key, so existing callers
work unchanged on XPU devices.

Previously `activation` declared only `cuda`, `rocm`, `metal` and `cpu`. On XPU
the kernel could not be loaded at all, forcing every consumer to carry a
separate eager fallback path.

Operators: `silu_and_mul`, `mul_and_silu`, `gelu_and_mul`, `gelu_tanh_and_mul`,
`fatrelu_and_mul`, `gelu_new`, `gelu_fast`, `gelu_quick`, `silu`, `gelu`,
`gelu_tanh`. Supported dtypes: `float32`, `float16`, `bfloat16`.

New sources live in `activation_xpu/`, following the layout of the existing CPU
backend. `tests/kernels/test_activation.py` is parameterized over the available
accelerator instead of hard-coding CUDA, so the existing suite runs unmodified
on XPU.

Depends on huggingface/kernels#752, which makes `kernel-builder` request IEEE
correctly rounded fp32 division for SYCL kernels.

## Correctness

Validated on Intel Arc Pro B60 with `torch 2.13.0+xpu` and oneAPI 2026.0.

The existing `tests/kernels/test_activation.py` passes in full on XPU (420
cases), including the gated operators that are compared against eager PyTorch
bit-for-bit with zero tolerance. Two cases were added to that file, covering
shapes and alignments that force a vectorized backend onto its scalar path, and
empty inputs. Both are device-parameterized like the rest of the suite.

## Performance

Measured on one Intel Arc Pro B60. Speedup is eager time divided by kernel time;
values above 1.00 favour the kernel. Gated operators use a `16384 x 16384`
input, element-wise operators `16384 x 8192`; the `4096` columns are the same
measurement at `4096 x 8192` and `4096 x 4096`.

| Operator | fp16 | bf16 | fp32 | fp16 (4096) | bf16 (4096) | fp32 (4096) |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| `silu_and_mul` | 1.66x | 1.66x | 1.68x | 1.80x | 1.67x | 1.67x |
| `mul_and_silu` | 1.66x | 1.66x | 1.68x | 1.66x | 1.67x | 1.67x |
| `gelu_and_mul` | 1.66x | 1.66x | 1.67x | 1.68x | 1.67x | 1.67x |
| `gelu_tanh_and_mul` | 1.87x | 1.64x | 1.71x | 1.97x | 1.69x | 1.72x |
| `fatrelu_and_mul` | 3.37x | 3.55x | 2.23x | 3.40x | 3.62x | 2.23x |
| `gelu_new` | 5.40x | 3.86x | 9.05x | 5.19x | 3.73x | 8.53x |
| `gelu_fast` | 6.20x | 4.48x | 10.51x | 6.01x | 4.35x | 10.00x |
| `gelu_quick` | 3.44x | 3.47x | 3.42x | 3.54x | 3.52x | 3.49x |
| `silu` | 0.98x | 0.99x | 0.98x | 1.02x | 1.00x | 1.00x |
| `gelu` | 0.98x | 0.99x | 0.97x | 1.01x | 1.01x | 1.00x |
| `gelu_tanh` | 1.18x | 1.05x | 1.02x | 1.15x | 1.01x | 0.95x |

Three regimes are visible:

* **Gated operators** avoid two `[n, d]` intermediates and cut memory traffic
  from `6nd` to `3nd`. `fatrelu_and_mul` gains more because its eager form
  additionally materializes a boolean mask.
* **Composite activations** have no native PyTorch operator and expand into
  eight or more element-wise launches in eager mode, so fusing them gives the
  largest speedups.
* **Single-operator activations** read and write the same amount of data as the
  corresponding PyTorch operator and are therefore on par with it. They exist
  for interface completeness, so callers need no device-dependent branch.

### Memory

Peak memory allocated while producing the result, via
`torch.xpu.max_memory_allocated`. The output tensor is included and the input
excluded, so both paths are charged for what they produce. The ratio is
independent of dtype and shape; absolute figures are `float16` at
`16384 x 16384` (gated) and `16384 x 8192` (element-wise).

| Operator | eager | kernel | reduction |
| --- | ---: | ---: | ---: |
| `silu_and_mul`, `mul_and_silu`, `gelu_and_mul`, `gelu_tanh_and_mul`, `fatrelu_and_mul` | 512 MiB | 256 MiB | 2.00x |
| `gelu_quick` | 512 MiB | 256 MiB | 2.00x |
| `gelu_new` | 768 MiB | 256 MiB | 3.00x |
| `gelu_fast` | 1024 MiB | 256 MiB | 4.00x |
| `silu`, `gelu`, `gelu_tanh` | 256 MiB | 256 MiB | 1.00x |

The fused kernels never allocate beyond the output tensor. Eager `gelu_fast`
peaks at four times that because three intermediates are live simultaneously.

## Build

Both XPU variants offered by `kernel-builder` build cleanly and pass the
`manylinux_2_28` and Python `abi3` compatibility checks:

```
nix build '.#redistributable.torch213-cxx11-xpu20260-x86_64-linux'
nix build '.#redistributable.torch212-cxx11-xpu20253-x86_64-linux'
```

## Limitations

* Inputs must be contiguous, matching the CUDA backend.

### DIFF
--- a/activation/activation_xpu/activation_kernels.cpp
+++ b/activation/activation_xpu/activation_kernels.cpp
@@ -1,0 +1,422 @@
+#include <sycl/sycl.hpp>
+
+#include <ATen/ATen.h>
+#include <ATen/DeviceGuard.h>
+#include <ATen/Dispatch.h>
+#include <c10/xpu/XPUStream.h>
+#include <torch/torch.h>
+
+#include <algorithm>
+#include <cstdint>
+
+#include "activation_xpu.hpp"
+
+namespace activation_xpu {
+
+namespace {
+
+// Aligned pack of `vec_size` elements, used to widen global memory accesses so
+// that a single work item issues one wide load/store instead of several narrow
+// ones. `sycl::vec` cannot be used here because `at::Half` and `at::BFloat16`
+// are not valid SYCL vector element types.
+template <typename scalar_t, int vec_size>
+struct alignas(sizeof(scalar_t) * vec_size) AlignedVec {
+  scalar_t val[vec_size];
+};
+
+// Widest access that is still a natural vector width on Intel GPUs (128 bit).
+template <typename scalar_t>
+constexpr int kVecSize = 16 / sizeof(scalar_t);
+
+bool is_vec_aligned(const void* ptr, int64_t bytes) {
+  return reinterpret_cast<uintptr_t>(ptr) % bytes == 0;
+}
+
+// Activation and gating kernel. Each work-group handles one token, the work
+// items in the group stride over the `d` elements of that token.
+template <typename scalar_t, typename act_fn_t, bool act_first>
+struct ActAndMulKernel {
+  ActAndMulKernel(scalar_t* out, const scalar_t* input, int64_t d,
+                  act_fn_t act_fn)
+      : out_(out), input_(input), d_(d), act_fn_(act_fn) {}
+
+  void operator()(sycl::nd_item<1> item) const {
+    const int64_t token_idx = item.get_group(0);
+    const int64_t local_size = item.get_local_range(0);
+    const scalar_t* input_row = input_ + token_idx * 2 * d_;
+    scalar_t* out_row = out_ + token_idx * d_;
+    for (int64_t idx = item.get_local_id(0); idx < d_; idx += local_size) {
+      const scalar_t x = input_row[idx];
+      const scalar_t y = input_row[d_ + idx];
+      out_row[idx] = act_first ? act_fn_(x) * y : x * act_fn_(y);
+    }
+  }
+
+ private:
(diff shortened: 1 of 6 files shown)


jiqing-feng:
https://github.com/huggingface/kernels/pull/752 merged. This PR is ready to be reviewed.

danieldk (approved):

PR #1074, opened 2026-08-13
fla: fix relative import depth in triton_ascend backend

jiqing-feng:
## Summary

`TritonAscendUtilsBackend.is_available()` does `from ...utils import IS_NPU`,
but the backend lives at `fla.ops.utils.backends.triton_ascend`, so three levels
up resolves to the non-existent `fla.ops.utils.utils`.

This is not dormant on non-NPU hardware. `BackendRegistry.register()` calls
`_update_active_backend()` -> `can_use()` -> `is_available()` eagerly at import
time, so the broken import raises on every platform, taking down any op routed
through `chunk_local_cumsum` — `chunk_gated_delta_rule` among them.

The two `chunk_*_cumsum` methods reach the sibling `cumsum` module by absolute
path through a relative prefix; switched to plain relative imports, which
`AGENTS.md` asks for anyway.

This is a vendoring bug, not an upstream one — `fla-org/flash-linear-attention`
uses absolute imports here and is unaffected. The depth was miscounted when the
snapshot was converted to the relative imports `AGENTS.md` requires, so there is
nothing to report upstream.

## Validation

`chunk_gated_delta_rule` imports and runs after the fix; it raised on import
before. Checked on Intel Arc Pro B60, `torch 2.13.0+xpu`, but the failure is not
backend specific.


jiqing-feng:
Hi @danieldk 
Would you please review this PR? Thx!

danieldk (approved):

PR #1075, opened 2026-08-13
triton-kernels: fix Intel opt-flag heuristics and make matmul_ogs_torch usable off CUDA

jiqing-feng:
## Summary

**Intel opt flags.** `opt_flags_intel.py` was copied from the NVIDIA path and kept two constants that describe NVIDIA hardware:

- `block_k` was hardcoded to `min(128, ...)` instead of being derived from the cacheline size. An Xe cacheline is half as wide, so operand tiles were twice the size the hardware wants. Deriving it the NVIDIA way needs the operand dtypes, hence the signature change.
- `num_warps = block_m * block_n // 4096` assumes a 32-lane warp. An Xe sub-group is SIMD16, so the same tile needs twice as many warps to keep per-lane register pressure in range; a `tl.dot` sweep puts the optimum at `// 2048`.

Also allows `block_n` as a tuning constraint, so the Intel path can be swept from the outside like `block_m` and `block_k` already can. Only the `backend == "xpu"` branch is touched; CUDA and HIP are unaffected.

**`matmul_ogs_torch`.** The reference implementation failed with its own documented defaults: `device` defaulted to the string `"cuda"`, raising "Torch not compiled with CUDA enabled" on a non-CUDA build (now defaults to `x.device`, a no-op on CUDA), and the fallback `round_x = lambda x: x` takes one argument while the call site passes two, so any caller not supplying `round_x` got a `TypeError` on every backend. Supersedes #1076.

## Validation

bf16 on Intel Arc Pro B60, one card, `matmul_ogs` before vs after:

| shape | speedup |
| --- | ---: |
| MoE, 4096 tokens / 32 experts / top4 | 9.5x |
| MoE, 2048 tokens / 8 experts / top2 | 8.6x |
| dense 4096³ | 4.2x |

Numerics unchanged: max relative error against `matmul_ogs_torch` stays at the same order as before. The `matmul_ogs_torch` fix is what makes that check runnable off CUDA in the first place.

A gap to oneDNN remains, but it is in the Triton XPU backend, not in these heuristics.


danieldk (approved):
PR looks good, also changed the tests so that we can run them in CI.

PR #1084, opened 2026-08-15
sgl-flash-attn3: rename #include-d instantiation .cu files to .cuh to avoid double compilation

### LINKED ISSUE #1032: Propagate FA3 build fix: rename `#include`-d `.cu` files to `.cuh` to avoid double-compilation
Tracking issue to propagate the build optimization landed in Flash Attention 3 (huggingface/kernels-community#999) to other kernel sources in this org when relevant.

## Background

FA3 (and some other kernels) have a pattern where `.cu` files are `#include`-d inside other `.cu` files. Because the CMake rule compiles every `.cu` file, the included files end up compiled **twice** — once directly, once through the file that includes them. This inflates both compilation time and binary size.

The core issue is that `#include` semantics are "literally paste the file contents here," but that breaks down once build rules are applied to files by extension.

## The fix

Rename any `.cu` file that is only meant to be `#include`-d into another translation unit to `.cuh`, so it's treated as a header/dependency of the kernel component rather than a separately-compiled source file.

## Impact (FA3, PR #999)

- Substantially faster compilation.
- Binary size reduction, e.g. **821 MiB → 550 MiB** for CUDA 13.

We could apply the same to `sgl-flash-attn3` and `vllm-flash-attn3`.

## Ask

Audit other kernels in `kernels-community` for the same `#include`-d-`.cu` pattern and apply the equivalent rename where relevant.

---
Requested by Sayak Paul - [Slack thread](https://huggingface.slack.com/archives/C090JN2P8NB/p1782833119913869?thread_ts=1782833119.913869&cid=C090JN2P8NB) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-07-16T10-45-37-234Z_85aa8964-f546-4d38-8acd-00ac8da29e4f.jsonl)

bikrammajhi:
Closes #1032

## What
Propagates the FA3 build fix (already applied to `vllm-flash-attn3` in #998) to `sgl-flash-attn3`.

In `flash-attn/instantiations/`, the template-instantiation files named `*_hdim*_*_{split,softcap,hdimall,hdimdiff}_sm*_*.cu` are **aggregators**: they `#include` the base instantiation files (e.g. `flash_fwd_hdim128_bf16_sm90.cu`) to avoid recompiling the heavy templates per variant. Those base files were also listed in `build.toml` `src`, so the kernel-builder compiled them twice: once directly and once through each aggregator that includes them.

This renames the 300 base files that are `#include`-d by aggregators from `.cu` to `.cuh`, so the builder treats them as headers rather than standalone translation units:

- `sgl-flash-attn3/flash-attn/instantiations/*.cu` → `*.cuh` (300 files, pure renames)
- the 100 aggregator `.cu` files now `#include "...cuh"` instead of `"...cu"`
- `sgl-flash-attn3/build.toml` `src` entries updated `.cu` → `.cuh` accordingly (100 `.cu` + 300 `.cuh`, same as the vllm fix which keeps renamed files in `src`)

`flash_fwd_combine.cu` and `flash_prepare_scheduler.cu` are unaffected (they are compiled directly and never `#include`-d).

## Verification
- No `.cu` file in `instantiations/` still `#include`s a `.cu` file.
- All `.cuh` files referenced from aggregators and `build.toml` exist on disk; all `.cu`/`.cuh` on disk are referenced in `build.toml`.
- `build.toml` parses (tomllib) with 100 `.cu` + 300 `.cuh` instantiation entries.


bikrammajhi:
Thank you for taking the time to review this. I appreciate the feedback and will keep it in mind for future contributions.

PR #1087, opened 2026-08-20
 fix how we handle uploads during staging.

sayakpaul:
Fix situations like https://github.com/huggingface/kernels-community/actions/runs/31891391510/job/95028144576 where staging is trying to upload to the vendor org where it should have uploaded to the kernels-staging org.


drbh (approved):
nice fix! lgtm!

PR #1090, opened 2026-08-21
deep-gemm: smoke tests

sayakpaul:


deep-gemm/tests/test_kernels_ci.py
```diff
@@ -0,0 +1,151 @@
+# Smoke tests for the kernels-community CI runner (`nix run .#ci-test`, which
```
sayakpaul, line 1: Only suite that will run on our CI. Keeping them lightweight is a necessity.

deep-gemm/tests/generators.py
```diff
@@ -3,14 +3,21 @@
...
-from deep_gemm.utils import (
-    align, ceil_div,
-    per_token_cast_to_fp8, per_channel_cast_to_fp8, per_block_cast_to_fp8,
-    per_token_cast_to_fp4, transpose_packed_fp4,
-    get_mk_alignment_for_contiguous_layout,
-    set_mk_alignment_for_contiguous_layout
-)
+import kernels
+
+deep_gemm = kernels.get_kernel("kernels-community/deep-gemm", version=2)
```
IlyasMoutawwakil, line 8: does the ci testing load the hub version or builds its own ?

deep-gemm/torch-ext/deep_gemm/__init__.py
```diff
@@ -770,9 +771,11 @@ def _find_cuda_home() -> str:
...
             cuda_home = "/usr/local/cuda"
-            if not os.path.exists(cuda_home):
-                cuda_home = None
-    assert cuda_home is not None, "Could not find CUDA installation"
+    # Only the runtime JIT compiler needs a CUDA toolkit. The cuBLASLt GEMMs, the
+    # runtime configuration surface and the layout fast paths compile nothing, so
+    # a missing toolkit must not fail here -- `init()` only records the path.
+    # Return the conventional location and let the JIT compiler report the
+    # missing `nvcc` if it is ever reached.
     return cuda_home
```
IlyasMoutawwakil, line 779: hmm i'm not sure but hope this doesn't break anything

IlyasMoutawwakil (approved):
LGTM

deep-gemm/tests/generators.py
sayakpaul: Hub version. Testing is done after the build and upload is done.

deep-gemm/torch-ext/deep_gemm/__init__.py
sayakpaul: It shouldn't because the JIT compiler will report the error on its own.

sayakpaul:
`testing` module isn't typically used in user-facing applications. So, the failures are acceptable IMO.

PR #1091, opened 2026-08-21
weathernext2-banded-attention: add WeatherNext 2 banded mesh-attention kernel

kashif:
Inference Triton kernel for WeatherNext 2's mesh attention, for the port in huggingface/transformers#47874.

The mesh nodes are ordered by reverse Cuthill-McKee, so the adjacency is banded: a block of nodes only reaches itself and its two neighbours. The in-tree path materialises the three neighbouring key/value blocks and hands a `[blocks, 1, block, 3 * block]` mask to sdpa. This walks the neighbours in place instead, streams the mask, and skips tiles the band never reaches.

1.73x end to end on a forecast step with the released 1 degree checkpoint on an H100 (223.7 -> 129.1 ms). Profiling put attention at 21% of the step and the graph scatter at 0.5%, so this is the only part worth a kernel.

Upstream runs this attention in fp32, so `precision` picks how `tl.dot` treats fp32 inputs: `tf32` by default, `tf32x3`, or `ieee`. In `ieee` it matches sdpa to 2e-6; in `tf32` the forecast moves 2.4e-3 of each field's spatial spread against the JAX reference, versus 7e-4 for the shipped path. Forward only, so it should be registered inference-only.

Two things to flag:
- ROCm is declared but untested, no AMD GPU to hand. It matters more there than on CUDA: the fp32 fallback PyTorch picks is `fmha_cutlassF_f32_aligned`, which has no ROCm equivalent, and the math fallback materialises the full score matrix.
- The fast path needs the raw banded mask, so it needs a companion transformers change. Without it the layer falls back to gather-plus-sdpa, correct but unaccelerated.

Structure and idioms follow `esmfold2-trimul` (`input_precision`, the `PRECISION` constexpr, the accumulator form of `tl.dot`).


kashif:
Tested on ROCm now, so the "untested" caveat in the description is stale. gfx1150, torch 2.13.0+rocm7.2, triton 3.7.1: all 7 tests pass, matching sdpa to 6.9e-07.

Speed and memory against sdpa there, sweeping block size:

| block_size | sdpa ms / GiB | kernel ms / GiB | speed | memory |
|---|---|---|---|---|
| 256 | 2.56 / 0.07 | 2.83 / 0.05 | 0.90x | 0.71x |
| 512 | 9.11 / 0.18 | 12.22 / 0.07 | 0.75x | 0.39x |
| 1024 | 34.59 / 0.56 | 48.45 / 0.10 | 0.71x | 0.18x |

Slower per step on that device, but the memory advantage widens sharply with block size, since sdpa grows with `block_size**2` and the kernel stays near flat. Torch also warns that memory-efficient attention on AMD is still experimental, so sdpa is on the math path. gfx1150 is integrated RDNA 3.5 with no tf32, at a tenth of the model's real block size, and the autotune configs were picked on an H100, so I would not read those timings as representative of CDNA.

Also fixed since opening:
- launches now sit inside `device_context(...)`, matching liger-kernels and finegrained-fp8. Without it a shard placed on `cuda:1` by `device_map="auto"` would have been launched against `cuda:0`.
- the layer refuses the fast path when autograd is live. The kernel's output has no `grad_fn`, so `backward()` would have succeeded while every parameter upstream of attention silently got nothing.
- registered in `KERNEL_SOURCE_MAPPING`.

`check-public-api` fails because `scripts/check_public_api.py` treats added symbols as a change and has no new-kernel exemption, and `dispatch` 403s because fork PRs cannot dispatch workflows. Both need a maintainer.

sayakpaul:
kernel-bot build-and-stage weathernext2-banded-attention

danieldk (approved):

PR #1092, opened 2026-08-21
feat: prefer recipe ported einops

drbh:
*updated to match latest changeset

this pr is an example of using the kernel porting tool to port the upstream einops into the kernel builder format

first we clone the upstream source*

```bash
# first clone the repo locally
git clone https://github.com/arogozhnikov/einops.git /tmp/einops-src

# next we pin to a specific commit (same used in the recipe)
git -C /tmp/einops-src checkout ad2c8d67d08cbebb45e1aff2be21dc88635fc695
```

next we 
```bash
# run the tool directly from the branch (could later be built into the cli tool)
nix run 'github:huggingface/kernels/pull/776/head#kernel-port' -- \
einops/port/port.kdl \ # the recipe to port
--dir /tmp/einops-src \ # the location of the source we cloned
--out ~/Projects/kernels-community/einops # the output location
```

running the command will show how each line of the recipe impacted the transformation and outputs all of the files added and deleted in the process.

```
[line   4] source              verified https://github.com/arogozhnikov/einops.git @ ad2c8d67d08c
[line   9] prune               removed 57 file(s), kept 27
[line  12] delete              removed 1 file(s)
[line  15] move                moved 8 file(s) to "tests"
[line  20] move                moved 18 file(s) to "torch-ext/einops"
[line  25] relativize_imports  rewrote 7 import(s) in 4 file(s)
[line  30] ensure_import       added 1 import(s) in 1 file(s)
[line  36] remap_module        rewrote 7 import(s) in 6 file(s)
[line  42] relativize_imports  rewrote 7 import(s) in 6 file(s)
[line  47] kernelize_imports   kernelized 26 import(s) in 8 file(s)
[line  54] replace             1 replacement(s) in 1 file(s)
[line  59] replace             1 replacement(s) in 1 file(s)
[line  64] replace             7 replacement(s) in 7 file(s)
[line  70] replace             1 replacement(s) in 1 file(s)
[line  75] manifest            wrote build.toml (noarch)
```

`check-ports.yaml` reruns the port in ci and fails if `einops/src` does not match the output, so the generated tree is still reviewable but cannot drift from the recipe. prs only check the ports they touch

note: `flake.nix` is not generated so it stays at `einops/flake.nix` and points nix at `./src`, `build.yaml` already handles this layout (https://github.com/huggingface/kernels-community/pull/1109)


drbh:
for completeness, we can bump the commit hash from `ad2c8d67d08cbebb45e1aff2be21dc88635fc695` to the latest head `e0d5eb4fd535945ff65d309206d0f1754a926821`

and trying to rerun the command shows

```
nix run 'github:huggingface/kernels/pull/776/head#kernel-port' -- .ports/einops/port.kdl --dir /tmp/einops-src --out einops
[line  14] source              verified https://github.com/arogozhnikov/einops.git @ e0d5eb4fd535
[line  18] prune               removed 56 file(s), kept 27
[line  19] delete              removed 10 file(s)
[line  21] move                moved 17 file(s) to "torch-ext/einops"
error: recipe line 25: relativize_imports: expected exactly 7 change(s) but made 13 - upstream drifted; review the new rewrites and update changes=
```

at this point I simply asked and llm to fix the recipe so the port works correctly and copy pasted the command and error output above. 

the exact output of this process can be seen in https://github.com/huggingface/kernels-community/pull/1093

sayakpaul:
I think we should have a single standard to start this. I would prefer if we required users to specify the source and the commit in the recipe itself instead of having them locally clone and checkout to a particular branch first. Or, that should be a better default behaviour, IMO.

> 9 of the 15 files in the overlay are tests we added that are not ported from upstream

We should port tests, too. Then it will be down on us to make modifications there e.g., for running specific tests on our CI, etc. Cc: @danieldk here as well.

sayakpaul:
Going to run the commands from the PR description and share my findings here as well.

einops/src/port-provenance.json
```diff
@@ -0,0 +1,12 @@
+{
```
sayakpaul, line 1: Is this automatically generated by https://github.com/huggingface/kernels/pull/776?

.ports/einops/port.kdl
```diff
@@ -0,0 +1,60 @@
...
+
+// array_api is a submodule upstream leaves out of the package init; the kernel
+// exposes it, since a Hub consumer cannot import einops.array_api by name.
+replace \
+    in="torch-ext/einops/__init__.py" \
+    count=1 \
+    find="from .packing import pack, unpack\n" \
+    with="from .packing import pack, unpack\nfrom . import array_api\n"
+
+// The rewritten tests and the flake.
```
sayakpaul: We should not have to rewrite the tests much though. I think in the tests we would like to ensure that they use `get_kernel()` which already seems to be the case. So, I think we should rather provide similar instructions in the recipe to account for this.

`pytest.mark.kernels_ci` should probably be left to us because it's better if we decided which tests are lightweight and important enough to run on our CI.

.ports/einops/port.kdl
```diff
@@ -0,0 +1,60 @@
+recipe version=1
```
sayakpaul: I think we also decided to not maintain any top-level `.ports`?

.ports/einops/overlay/tests/test_other.py
```diff
@@ -0,0 +1,307 @@
...
+                x = numpy.ones(shape)
+                parse_shape(backend.from_numpy(x), pattern)
+
+
+_SYMBOLIC_BACKENDS = [
+    *collect_test_backends(symbolic=True, layers=False),
+    *collect_test_backends(symbolic=True, layers=True),
+]
+
+# tensorflow.keras needs special way to compile,
```
sayakpaul, line 186: Didn't expect this term to appear in 2026.

.ports/einops/port.kdl
```diff
@@ -0,0 +1,60 @@
...
+
+// einops
```
sayakpaul: Could we see an example where replace op is being used for torch op registration?

.ports/einops/port.kdl
sayakpaul: And also, I think tests should not be overlays no? Currently, we include them inside of the actual kernel source.

sayakpaul:
https://github.com/huggingface/kernels-community/pull/1100 -- used the tool on `flash-attn-ops`. Major changes are in a layer norm file and I provided some comments there.

einops/src/port-provenance.json
drbh: yep

.ports/einops/port.kdl
drbh: the tests actually come from upstream and the port was updated so we rewrite the import to use get_kernels, this added some more lines to the port - but now we do not have any tests in the overlay file.

however if we were to add tests that were not originally upstream and rewritable - they would belong in the overlays folder - since that should contains all of the code that is not upstream. the "kernel source" which is now in the `einops/src` file should not contain any code that is not output from the port too. we keep this clear separation so the `einops/src` only contains generated code (we never mix non generated code with the tools output)

.ports/einops/port.kdl
drbh: yes this has been removed in the lates and now is in `einops/port`

.ports/einops/overlay/tests/test_other.py
drbh: hahah yea feel outdated but only exists because its upstream here https://github.com/arogozhnikov/einops/blob/1f2268c55a25e1d422680a2f22a4c71f4faa7f3d/einops/tests/test_other.py#L188

.ports/einops/port.kdl
drbh: yes! I'll followup with a more meaningful example (rewrite torch ops) in the next example/pr. I stared with einops mostly to introduce the tool and keep things as small/low risk as possible.

sayakpaul:
@drbh can we build using the bot without the PR not having landed in `main`?

einops/port/port.kdl
```diff
@@ -0,0 +1,293 @@
...
+    count=0
+
+// array_api is a submodule upstream leaves out of the package init; the kernel
+// exposes it, since a Hub consumer cannot import einops.array_api by name.
+replace in="torch-ext/einops/__init__.py" count=1 \
+    find="from .packing import pack, unpack" \
+    with=#"""
+from .packing import pack, unpack
+from . import array_api
+"""#
```
danieldk: This seems quite verbose? It would be nicer to have something less literal and more declarative that is "for every import of the kernel of the shape `import kernel` replace by `kernel = get_kernel(...)` and every import of the shape `from kernel import foo` replace by `foo = kernel.foo`. All these full snippets are going to be very fragile for active upstream projects.

einops/port/port.kdl
```diff
@@ -0,0 +1,293 @@
...
+            __import__(module)
+            available.append(name)
+        except ImportError:
+            pass
+    return available
+"""#
+
+replace in="tests/test_einsum.py" count=1 \
+    find=#"""
+from einops.tests import collect_test_backends
```
danieldk: This seems quite verbose? It would be nicer to have something less literal and more declarative that is "for every import of the kernel of the shape `import kernel` replace by `kernel = get_kernel(...)` and every import of the shape `from kernel import foo` replace by `foo = kernel.foo`. All these full snippets are going to be very fragile for active upstream projects.

.ports/einops/port.kdl
sayakpaul: I agree. Those tests (that we modify heavily) should probably be reserved for just CI-level testing. Those are fine to have in overlays.

.github/workflows/check-ports.yaml
```diff
@@ -0,0 +1,43 @@
+name: Check Ports
+
+# A ported kernel's <kernel>/src is generated from <kernel>/port/port.kdl and
+# committed. Regenerate it here and fail if it drifted, so that a stale overlay
+# silently reverting an in-tree fix, a hand-edited build.toml field the manifest
+# op drops, or an overlay file edited without its generated twin are all loud.
```
sayakpaul: What do we consider to be a drift here?

.github/workflows/check-ports.yaml
```diff
@@ -0,0 +1,43 @@
...
+on:
+  push:
+    branches: [main]
+  pull_request:
+    branches: [main]
+    types: [opened, synchronize, reopened]
+    paths:
+      - "*/port/**"
+      - "*/src/**"
```
sayakpaul, line 20: So for kernels containing valid porting recipes, we expect their sources to be present under `src`, which is as expected 👍

.github/workflows/check-ports.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+      - uses: DeterminateSystems/nix-installer-action@ef8a148080ab6020fd15196c2084a2eea5ff2d25 # v22
+
+      # Each recipe pins its own upstream commit; the runner verifies the
+      # origin URL, HEAD and cleanliness of the checkout it is handed.
+      - name: Regenerate every port and compare
+        env:
+          # kernel-port still lives on the kernel-port-tool branch; point this at
+          # github:huggingface/kernels#kernel-port once that branch merges.
+          KERNEL_PORT: "nix run github:huggingface/kernels/kernel-port-tool#kernel-port --"
```
sayakpaul: Let's add a TODO here to update when the PR is eventually merged.

einops/src/port-provenance.json
```diff
@@ -0,0 +1,12 @@
...
+  "format": 1,
```
sayakpaul, line 2: Self TODO: review the provenance generation logic.

einops/port/port.kdl
```diff
@@ -0,0 +1,80 @@
+recipe version=1
```
sayakpaul, line 1: Okay so IIUC we didn't need any overlay in this case? I thought the flake.nix would go to the overlay directory? How are we handling that?

.github/workflows/check-ports.yaml
sayakpaul: Also, once this is shipped do we expect users to run with `nix run` or through the `kernel-builder` CLI?

scripts/check_ports.py
```diff
@@ -0,0 +1,157 @@
...
+    kernels = sys.argv[1:] or find_ports(root)
+    if not kernels:
+        print("No ported kernels found (no */port/port.kdl).")
+        return 0
+
+    print(f"Checking {len(kernels)} port(s) reproduce their committed tree:\n")
+    failed = {}
+    with tempfile.TemporaryDirectory() as tmp:
+        workdir = Path(tmp)
+        for kernel in kernels:
```
sayakpaul, line 158: Instead of check all the kernels, shouldn't we check only for the kernels that were changed / added in a PR? We could also add an option to check all kernels but by default that probably shouldn't be the case?

sayakpaul (commented):
Thanks, left some comments. I like the idea of validating a port, really cool!

.github/workflows/check-ports.yaml
drbh: drift here is just `einops/src` not matching what the recipe outputs. the cases this catches are an overlay file going stale and undoing a fix that was made directly in `src`, a manual edit to `build.toml` that the `manifest` op drops on the next run, and the recipe being edited without rerunning the port (`port-provenance.json` stores the recipe hash so that one shows up as a diff in the provenance file). updated the comment in the workflow to list these

.github/workflows/check-ports.yaml
drbh: > Let's add a TODO here to update when the PR is eventually merged.

done, also pinned the runner to a commit on the `kernel-port-tool` branch instead of the branch head so changes there dont affect open prs

> Also, once this is shipped do we expect users to run with nix run or through the kernel-builder CLI?

`nix run` for now since the tool only exists on the branch. once it merges I think we can expose it as a `kernel-builder` subcommand and update `KERNEL_PORT` in the workflow, the script just reads that env var so either works

einops/port/port.kdl
drbh: yea no overlay is needed for einops since everything comes from upstream. `flake.nix` is not generated so it stays at `einops/flake.nix` with `path = ./src` and only the tool output lives under `src/`. `build.yaml` already picks up `<kernel>/src/build.toml` when it exists

scripts/check_ports.py
drbh: yea good point, prs now run `check_ports.py --changed-since origin/main` which only checks the ports where `port/` or `src/` changed. pushes to `main` and manual runs still check everything

einops/port/port.kdl
sayakpaul: > flake.nix is not generated so it stays at einops/flake.nix 

Oh okay. Earlier, I think we wanted to keep the external files in the overlay. What is the general guidelines behind putting stuff under an `overlay` folder then?

scripts/check_ports.py
sayakpaul: Perfecto!

PR #1093, opened 2026-08-21
feat: bump pinned upstream and rerun port

drbh:
this pr is an example of bumping upstream, adjusting the recipe and rerunning the port tool to show how we can track upstream commits

*note the only changes that were manually made are in `.ports/einops/port.kdl` all other files were automatically generated


drbh:
closing for now as the underlying pr is changing enough that this should be reopened at a later time

PR #1095, opened 2026-08-24
sage-attention: sync sage and sage2

sayakpaul:
We have had quite a bit of staleness in `sage-attention` which is picking up pace again for running large video models on consumer GPUs. So, this PR:

* Syncs Sage and Sage2 (which is better overall) with upstream.
* Removes Sage3 related stuff because it lives in a separate directory in upstream (will open a separate PR for that).
   * Since Sage3 is meant for Blackwell only, it also simplifies builds.
* Bumps the versions as needed. 
* Marks some tests for CI (doing this because Sage will become more and more useful in the media generation world).

Check the results here: https://github.com/huggingface/diffusers/pull/14584

Cc: @asomoza


sage-attention/build.toml
```diff
@@ -1,10 +1,17 @@
...
 name = "sage-attention"
-version = 2
+version = 3
+edition = 5
 license = "Apache-2.0"
 backends = ["cuda"]
+upstream = "https://github.com/thu-ml/SageAttention.git"
 
 [general.cuda]
+# Load-bearing: the CUDA 12.6 toolkit compiles the sm89 FP8 attention kernels
```
danieldk: "load-bearing", ai;dr 🤣

ok, I did read, but I have no idea what is load-bearing about it.

danieldk (dismissed):

sage-attention/build.toml
sayakpaul:
```suggestion
# CUDA 12.6 toolkit compiles the sm89 FP8 attention kernels
```

sayakpaul: Simplified it a bit.

sayakpaul:
@danieldk need another approval.

sage-attention/build.toml
```diff
@@ -64,6 +64,7 @@ backend = "cuda"
 cuda-capabilities = ["8.0"]
 cuda-flags = [
     "-O3",
+    "-std=c++17",
```
danieldk: Torch 2.14 requires C++20, so all `std=c++17` lines should be removed.

For some weird reason, GitHub currently does not allow me to do a code suggestion 😢.

sage-attention/build.toml
```diff
@@ -88,36 +90,53 @@ src = [
...
-    "--expt-extended-lambda",
     "--use_fast_math",
-    "-DNDEBUG",
-    "-DEXECMODE=0",
+    "--threads=1",
+    "-Xptxas=-v",
+    "-diag-suppress=174",
 ]
 cxx-flags = [
+    "-g",
```
danieldk: Why add extra debug information? Probably remove all `-g` as well.

sayakpaul: This was needed to inspect some intermediate build failures when I was in between the port update process. Left over. Removing.

danieldk (approved):

PR #1096, opened 2026-08-24
[chore]: improve public api checks and error reporting

### LINKED ISSUE #1030: Improve API-change check: gate on version bump and allow additive-only changes
The public-API check currently flags any change to a kernel's public API, even when the change is purely additive (e.g. adding new layer versions / new functions without modifying existing signatures). It also doesn't take the `build.toml` version into account, so bumping the version doesn't clear the warning.

Proposed improvements to the check script:
- Detect whether the version in `build.toml` was changed as part of the PR.
- Enforce that the version only ever increments (and validate it's a proper bump).
- Only run the public-API-diff check when the version was **not** bumped — if the version was bumped, the API change is expected and shouldn't be flagged.
- Consider allowing additive-only changes (new functions/layers, no changes to existing signatures) to pass without complaint, so we don't need a version bump for every new exposure.

Context: this came up while merging additive layer-version PRs (#949, #995, #996) where the API-change check fired even though no existing function signatures changed.

---
Requested by Sayak Paul - [Slack thread](https://huggingface.slack.com/archives/C090JN2P8NB/p1783949128431509?thread_ts=1783949128.431509&cid=C090JN2P8NB) - [Agent trace](https://huggingface.co/buckets/huggingface/moon-bot-memory/tree/sessions/2026-07-16T07-48-58-721Z_7a0e6a13-2e73-402f-b6d0-346741693c6d.jsonl)

sayakpaul:
https://github.com/huggingface/kernels-community/pull/1095/ bumped the version but the API check still complained. This PR fixes that. Additionally, it broadens the coverage of potential suspects that warrant a version bump as laid down [here](https://huggingface.slack.com/archives/C090JN2P8NB/p1785827928772879). 

Closes https://github.com/huggingface/kernels-community/issues/1030


scripts/check_public_api.py
```diff
@@ -346,9 +375,21 @@ def _tree(items: list) -> None:
             print(f"     {'  ' if last else '| '} {d}")
 
 
+# An __all__ that only gains names is an addition, not a break: the entry is
+# keyed per module and would otherwise mark every new export as "changed".
+def _all_grew(base: str, head: str) -> bool:
+    try:
+        old, new = ast.literal_eval(base), ast.literal_eval(head)
```
danieldk:
```suggestion
        old = ast.literal_eval(base)
        new = ast.literal_eval(head)
```

Otherwise quick reading suggest it's a tuple-returning function.

scripts/check_public_api.py
```diff
@@ -346,9 +375,21 @@ def _tree(items: list) -> None:
             print(f"     {'  ' if last else '| '} {d}")
 
 
+# An __all__ that only gains names is an addition, not a break: the entry is
```
danieldk: The comment seems wrong? An all that gains names is breaking.

scripts/check_public_api.py
```diff
@@ -346,9 +375,21 @@ def _tree(items: list) -> None:
...
+# keyed per module and would otherwise mark every new export as "changed".
+def _all_grew(base: str, head: str) -> bool:
+    try:
+        old, new = ast.literal_eval(base), ast.literal_eval(head)
+    except (ValueError, SyntaxError):
+        return False
+    return isinstance(old, list) and isinstance(new, list) and set(old) <= set(new)
```
danieldk:
```
set(old) <= set(new)
```

Wouldn't this also return `True` for cases where the old set is the same as the new set?

scripts/check_public_api.py
```diff
@@ -346,9 +375,21 @@ def _tree(items: list) -> None:
...
+
+
+# Returns (breaking, additive): breaking covers removed and re-signed symbols,
+# additive covers newly exposed ones. Both require a version bump.
 def report(
     kernel: str, base: dict, head: dict, limit: int = 20, preview: int = 6
-) -> bool:
+) -> tuple:
```
danieldk: Tuple typing, especially because the return statement is somewhat hard to read.

scripts/check_public_api.py
```diff
@@ -346,9 +375,21 @@ def _tree(items: list) -> None:
...
+# keyed per module and would otherwise mark every new export as "changed".
+def _all_grew(base: str, head: str) -> bool:
+    try:
+        old, new = ast.literal_eval(base), ast.literal_eval(head)
+    except (ValueError, SyntaxError):
+        return False
+    return isinstance(old, list) and isinstance(new, list) and set(old) <= set(new)
+
+
+# Returns (breaking, additive): breaking covers removed and re-signed symbols,
```
danieldk: What does `re-signed` mean?

scripts/check_public_api.py
```diff
@@ -380,7 +421,11 @@ def report(
     if len(items) > limit:
         items = items[:limit] + [(f"... and {len(items) - limit} more", [])]
     _tree(items)
-    return True
+
+    grown = [
```
danieldk, line 462: Isn't there only one `__all__`? If so, why a list comprehension?

danieldk: Would also benefit from some comments. E.g. what is the difference between `added` and `grown`?

scripts/check_public_api.py
```diff
@@ -380,7 +421,11 @@ def report(
...
+        k for k in changed if k.endswith(" __all__") and _all_grew(base[k], head[k])
+    ]
+    return bool(removed or set(changed) - set(grown)), bool(added or grown)
```
danieldk: Hard to read.

I think e.g. `set(changed) - set(grown)` could be computed in a separate line with a descriptive variable name and possibly comment.

scripts/check_public_api.py
```diff
@@ -435,19 +480,57 @@ def main() -> int:
...
+        new_version = kernel_version(head_src)
+        if (
+            isinstance(old_version, int)
+            and isinstance(new_version, int)
+            and new_version > old_version
+        ):
+            print(f"     => version bumped {old_version} -> {new_version}")
+            continue
+
+        flavour = additive_safe(head_src)
```
danieldk, line 552: Seeing the call site, why is `additive_safe` stringly typed?

danieldk (commented):
Thanks! Added a bunch of comments.

scripts/check_public_api.py
sayakpaul: I didn't understand the suggestion here.

scripts/check_public_api.py
sayakpaul: Would you go with a set?

scripts/check_public_api.py
sayakpaul: What would you prefer? `Enum`?

scripts/check_public_api.py
sayakpaul: Changed signature. Changed to a (hopefully) better comment.

scripts/check_public_api.py
sayakpaul: `scattermoe` has more than one:

* `torch-ext/scattermoe/__init__.py `
* `torch-ext/scattermoe/kernels/__init__.py`

scripts/check_public_api.py
sayakpaul: I changed it to a `dict` which I believe is more readable and easier to parse cognitively.

scripts/check_public_api.py
sayakpaul: Changed to `Enum` and I think it reads more explicit.

sayakpaul:
@danieldk addressed your comments

sayakpaul:
@drbh / @danieldk a review on it will be appreciated.

drbh (approved):
lgtm! 

**might be best for Daniel to give another once over before merging in case any of the comments need follow up

sayakpaul:
@danieldk could I get another review?

scripts/check_public_api.py
```diff
@@ -401,7 +437,7 @@ def report(
         if len(keys) > preview:
             items.append((f"... and {len(keys) - preview} more", []))
         _tree(items)
-        return False, False
+        return {"breaking": False, "additive": False}
```
danieldk: Seems like a perfect case for a small dataclass?

sayakpaul: Done.

PR #1099, opened 2026-08-25
sage-blackwell: port SageAttention3 Blackwell FP4 attention

sayakpaul:
Tested using RTX-PRO-6000 (GB202, sm_120)

**Note**

The kernel supports 12.0 and 12.1 but the latter isn't in our arch list yet. So, only 12.0. See: https://github.com/huggingface/kernels/issues/720.

### DIFF
--- a/sage-blackwell/torch-ext/sage_blackwell/__init__.py
+++ b/sage-blackwell/torch-ext/sage_blackwell/__init__.py
@@ -1,0 +1,18 @@
+"""
+Copyright (c) 2025 by SageAttention team.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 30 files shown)


sayakpaul:
@asomoza, you should be able to test with Blackwell Sage now:  https://huggingface.co/kernels/kernels-staging/sage-blackwell

```py

attn = get_kernel("kernels-staging/sage-blackwell", revision="pr-1099")
```

Do test it and let me know.

sage-blackwell/torch-ext/sage_blackwell/__init__.py
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+from .api import sageattn3_blackwell
+
+__all__ = ["sageattn3_blackwell"]
```
danieldk, line 18: I like simple APIs 😁

danieldk (approved):
Let's goooo!

sage-blackwell/torch-ext/sage_blackwell/__init__.py
sayakpaul: Aye!

PR #1100, opened 2026-08-25
[don't merge] port flash-attn-ops using the porting tool

sayakpaul:
Similar to https://github.com/huggingface/kernels-community/pull/1092.

### DIFF
--- a/flash-attn-ops/torch-ext/flash_attn_ops/layer_norm.py
+++ b/flash-attn-ops/torch-ext/flash_attn_ops/layer_norm.py
@@ -7,17 +7,18 @@
 # The models we train have hidden dim up to 8k anyway (e.g. Llama 70B), so this is fine.
 
 import math
-from typing import List, Optional
+from typing import Optional, List
 
 import torch
 import torch.nn.functional as F
-import triton
-import triton.language as tl
 from torch import Tensor
 
+import triton
+import triton.language as tl
+
 from ._ops import add_op_namespace_prefix
+from .utils.torch import custom_fwd, custom_bwd
 from .utils.library import triton_op
-from .utils.torch import custom_bwd, custom_fwd
 
 
 def maybe_contiguous_lastdim(x):
@@ -39,17 +40,9 @@ def triton_autotune_configs():
     warp_size = 32
     if torch.cuda.is_available():
         warp_size = getattr(torch.cuda.get_device_properties(torch.cuda.current_device()), "warp_size", 32)
-        warp_size = getattr(
-            torch.cuda.get_device_properties(torch.cuda.current_device()),
-            "warp_size",
-            32,
-        )
     # Autotune for warp counts which are powers of 2 and do not exceed thread per block limit
-    return [
-        triton.Config({}, num_warps=warp_count)
-        for warp_count in [1, 2, 4, 8, 16, 32]
-        if warp_count * warp_size <= max_threads_per_block
-    ]
+    return [triton.Config({}, num_warps=warp_count) for warp_count in [1, 2, 4, 8, 16, 32]
+            if warp_count * warp_size <= max_threads_per_block]
     # return [triton.Config({}, num_warps=8)]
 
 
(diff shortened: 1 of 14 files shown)


flash-attn-ops/torch-ext/flash_attn_ops/layer_norm.py
```diff
@@ -7,17 +7,18 @@
 # The models we train have hidden dim up to 8k anyway (e.g. Llama 70B), so this is fine.
 
 import math
-from typing import List, Optional
```
sayakpaul, line 10: Quite a bit of rewrite going on in this file with the porting tool. Some of them are formatting-related, though.

flash-attn-ops/torch-ext/flash_attn_ops/layer_norm.py
```diff
@@ -1266,9 +1198,8 @@ def backward(ctx, dout, *args):
         assert dy.shape == x.shape
         if ctx.prenorm:
             dresidual = args[0]
-            dresidual = maybe_contiguous_lastdim(
-                dresidual.reshape(-1, dresidual.shape[-1])
-            )
+            dresidual = maybe_contiguous_lastdim(dresidual.reshape(-1, dresidual.shape[-1]))
+            assert dresidual.shape == x.shape
```
sayakpaul, line 1202: This one seems out of the ordinary.

flash-attn-ops/torch-ext/flash_attn_ops/layer_norm.py
```diff
@@ -39,17 +40,9 @@ def triton_autotune_configs():
     warp_size = 32
     if torch.cuda.is_available():
         warp_size = getattr(torch.cuda.get_device_properties(torch.cuda.current_device()), "warp_size", 32)
-        warp_size = getattr(
-            torch.cuda.get_device_properties(torch.cuda.current_device()),
-            "warp_size",
-            32,
-        )
```
sayakpaul, line 46: Removes duplicate. So, that is a nice side-effect!

flash-attn-ops/torch-ext/flash_attn_ops/layer_norm.py
drbh: yea I think we must have ran a formatter and made some small changes manually in the past. its kinda nice that this surfaced those divergence

flash-attn-ops/torch-ext/flash_attn_ops/layer_norm.py
drbh: I think we must have run some formatter at some point since I think the changed version is what matches upstream on the pinned commit https://github.com/Dao-AILab/flash-attention/blob/b02b07e1a10238fe12831b80a8937ed59b1353a5/flash_attn/ops/triton/layer_norm.py#L1196

flash-attn-ops/torch-ext/flash_attn_ops/layer_norm.py
drbh: yea thats a nice catch and this aligns with upstream too https://github.com/Dao-AILab/flash-attention/blob/b02b07e1a10238fe12831b80a8937ed59b1353a5/flash_attn/ops/triton/layer_norm.py#L37

drbh:
this was a great exploration and its nice that it uncovered some divergences with upstream! its a good argument for us to use the tool on flash-ops once stabilized (the only thing this PR would need to change is the location of the port folder) - so we can either close this PR and reopen or make those changes to this PR

flash-attn-ops/torch-ext/flash_attn_ops/layer_norm.py
sayakpaul: But I would argue that it makes it difficult to catch the actual changes significantly more, especially on the existing kernels that we maintain in `kernels-community`. It becomes difficult to validate if those changes are purely formatting or if we should validate them with additional testing.

PR #1105, opened 2026-08-26
feat: add utilities to ping transformers maintainers on Slack for concerned kernels

sayakpaul:
Ad discussed internally.

We also now check in our CI if the kernels have an ownership status. They must either be owned or not owned. For a new kernel that is getting to `kernels-community`, one of the core maintainers anyway has to review and approve. So, they can take a decision if a kernel needs ownership. 

I have kept kernels like flash attention out of this list because that is general and the maintainers are responsible for it. Some kernels were intentionally left out because we don't see much activity there.


drbh (dismissed):
functionality looks good

nit: the changes are a bit verbose and it may be helpful for  maintenance to reduce the lines and remove things like `_comment` from the json file just to keep the changeset as small and targeted as possible

feel free to merge or clean up and then merge

sayakpaul:
@drbh yeah. Let me simplify a bit tomorrow and ask for another review.

sayakpaul:
@drbh sorry for the re-ping but I think I have considerably simplified to keep it less prone to maintenance burden and the changeset significantly more easily reviewable. Could you please take another look?

drbh (dismissed):
Nice! Huge improvement, much more simple and maintainable. Thanks for the changes 🙏

drbh (approved):

PR #1079, opened 2026-08-14
causal-conv1d: add conv1D kernel for XPU

kaixuanliu:

PR #1106, opened 2026-08-26
feat: Port causal conv1d

drbh:
this pr refers porting causal conv1d


kaixuanliu:
Hi @drbh, heads-up on a conflict: #1079  (merged last week, 1a0b434) added a SYCL/XPU backend to causal-conv1d, but this PR's base predates that merge — the recipe here has manifest backends="cuda" and no SYCL sources, so regenerating src would drop the XPU backend.

Since XPU has no upstream equivalent in Dao-AILab/causal-conv1d, prune keep=... won't preserve it. So can you help fold it into the recipe? Thx!

drbh:
closing until the first port is landed since the porting language is in flux until merged.

thanks for the heads up @kaixuanliu, I will make sure to keep this in mind when reopening the port pr

PR #1108, opened 2026-08-27
feat: Port flash attn2

drbh:
this pr prefers porting flash attn2


danieldk:
I think this needs to be rebased to avoid reintroducing the wild pointer issue.

drbh:
closing until the first port is landed since the porting language is in flux until merged

PR #1112, opened 2026-08-27
(vllm-)flash-attn3: fix version bound

danieldk:
Seems to have broken with the automatic update.


sayakpaul (approved):
Just precision issues 🤣

PR #1118, opened 2026-08-28
CI: vouch transformers members differently

danieldk:
Seems like automatic vouching of transformers requires a PAT, which is too much of a security risk. Instead, write a small script and add current members.


.github/VOUCHED.td
```diff
@@ -1,24 +1,60 @@
 # Vouched contributors for kernels-community
 # One GitHub handle per line, sorted alphabetically.
 # Denounce with a leading hyphen: -username reason
+3outeille
 Abdennacer-Badaoui
-IlyasMoutawwakil
```
danieldk, line 5: Not really removed, just UNIX-sorted, see below.

.github/scripts/sync-vouched-transformers.sh
```diff
@@ -0,0 +1,41 @@
...
+
+REPO_ROOT="$(git rev-parse --show-toplevel)"
+VOUCHED="$REPO_ROOT/.github/VOUCHED.td"
+TEAMS="transformers transformers-core-maintainers"
+
+members="$(mktemp)"
+trap 'rm -f "$members"' EXIT
+
+for team in $TEAMS; do
+  gh api --paginate "orgs/huggingface/teams/$team/members" --jq '.[].login' >>"$members"
```
sayakpaul, line 23: Does it not need any token? Can we check from an unauthenticated `gh` cli?

.github/scripts/sync-vouched-transformers.sh
```diff
@@ -0,0 +1,41 @@
+#!/usr/bin/env bash
```
sayakpaul, line 1: And how is this run on an interval basis or are we expected to run this script manually and open a PR with changes?

sayakpaul (approved):
Left some questions.

.github/scripts/sync-vouched-transformers.sh
danieldk: Yes, indeed. I don't think there is another way unless we want to dedicate a PAT to this.

.github/scripts/sync-vouched-transformers.sh
danieldk: See the comment above

```
Requires: gh (authenticated with read access to the huggingface org).
```

PR #1119, opened 2026-08-28
Set the correct repo-id for `mlx-rmsnorm` and `sgl-flash-attn3`

danieldk:


sgl-flash-attn3/build.toml
```diff
@@ -9,7 +9,7 @@ backends = ["cuda"]
 minver = "12.8"
 
 [general.hub]
-repo-id = "sgl-project/sgl-flash-attn3"
+repo-id = "kernels-community/sgl-flash-attn3"
```
sayakpaul, line 12: @adarshxs we should really be doing these updates to SGL project, instead. Can we ensure that?

danieldk: I think for now we should keep it as is, a lot of existing versions already depend on `kernels-community`.

danieldk: I think we could think about migrating once the Torch Stable ABI has been merged, because then we can keep the `kernels-community` flavor around and simply not update it anymore without huge breakage. But as it currently stands, we cannot even upload the stable ABI version without this PR merged.

danieldk: @sayakpaul good to merge?

sayakpaul (approved):

PR #1120, opened 2026-08-28
ggml-quantization: add llama.cpp GGUF quantization kernels for Metal

SunMarc:
# What does this PR do ?

This PR adds kernels from llamacpp related to quantization for metal only.

### DIFF
--- a/ggml-quantization/build.toml
+++ b/ggml-quantization/build.toml
@@ -1,0 +1,46 @@
+# One backend per `[kernel.*]` section, all implementing the same two entry points declared in
+# torch-ext/torch_binding.h. Adding Metal means adding a section and a `gguf_metal/` directory —
+# no change to the bindings' schema or to the Python API.
+#
+# `vendor/` is a pinned subset of llama.cpp (revision in vendor/UPSTREAM); refresh it with
+# `python vendor.py --rev <sha>`. It holds upstream's Metal kernels as they ship -- no local copy is
+# generated from them.
+
+[general]
+name = "ggml-quantization"
+version = 1
+license = "MIT"
+# Metal only for now. The cuda kernel and its build section were dropped rather than kept building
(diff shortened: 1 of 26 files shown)


SunMarc:
cc @sayakpaul

ggml-quantization/build.toml
```diff
@@ -0,0 +1,46 @@
...
+[general]
+name = "ggml-quantization"
+version = 1
+license = "MIT"
+# Metal only for now. The cuda kernel and its build section were dropped rather than kept building
+# untested -- `git log -- gguf_cuda` has them, and `vendor/` still carries ggml's cuda sources.
+backends = ["metal"]
+
+[general.hub]
+repo-id = "kernels-community/ggml-quantization"
```
sayakpaul, line 18: Okay for now but should we have this under the GGML org?

sayakpaul (dismissed):
I think it might be better to have this with https://github.com/huggingface/kernels-community/pull/1124 in a single PR.

ggml-quantization/build.toml
SunMarc: hmmm for now, i think it is better to have it here as it will be easier to maintain no ?

SunMarc:
Since the topk is a different package not related to ggml, i prefer to keep it seperate

ggml-quantization/build.toml
sayakpaul: GGML is HF :) So, even kernels-community is easier to maintain, we should be cognizant about growing kernels out of that single HF org. So, I think we should consider it.

SunMarc: I don't mind moving it to ggml but maybe let's see if this takes off or not before moving them to ggml org ? Both are fine to me cc @LysandreJik if you have a preference

sayakpaul:
@SunMarc let's fix the https://github.com/huggingface/kernels-community/actions/runs/33611755864/job/100188297542?pr=1120

sayakpaul (approved):

PR #1121, opened 2026-08-28
ggml-gated-delta-net: add llama.cpp gated delta net kernel for Metal

SunMarc:
# What does this PR do ?

This PR adds the gated-delta-net kernel (for qwen3.5) from llamacpp for metal only.


SunMarc:
cc @sayakpaul

sayakpaul:
@SunMarc I have triggered the workflow with the fixes again. Let's keep an eye

sayakpaul (approved):

PR #1122, opened 2026-08-28
ggml-attn: add llama.cpp flash-attention kernels for Metal

SunMarc:
## What does this PR do ? 

This PR adds the attn kernels from llama cpp for metal only.


SunMarc:
cc @sayakpaul

sayakpaul (approved):

Issue #1085, opened 2026-08-18
flash-attn2 v3 (stable ABI): backward fails for GQA/MQA


qgallouedec:
The v3 stable-ABI build runs the forward pass fine, but any backward with `num_kv_heads != num_heads` raises. MHA (`num_kv_heads == num_heads`) is unaffected, and the v2 `cxx11` build is fine on the same torch, so this looks specific to the stable-ABI build rather than to a torch version.

```python
import sys
import torch
from kernels import get_kernel


def check(version):
    flash_attn = get_kernel("kernels-community/flash-attn2", version=version)
    print(version, "->", sys.modules[flash_attn.__name__].__file__.split("/build/")[-1].split("/")[0])
    for num_heads, num_kv_heads in ((8, 8), (8, 2)):
        q = torch.randn(1, 8, num_heads, 64, dtype=torch.bfloat16, device="cuda", requires_grad=True)
        k = torch.randn(1, 8, num_kv_heads, 64, dtype=torch.bfloat16, device="cuda", requires_grad=True)
        v = torch.randn(1, 8, num_kv_heads, 64, dtype=torch.bfloat16, device="cuda", requires_grad=True)
        flash_attn.flash_attn_func(q, k, v, causal=True).sum().backward()
        print(f"  q={num_heads} kv={num_kv_heads}: backward ok")


check(2)  # torch212-cxx11-cu130-x86_64-linux: both fine
check(3)  # torch-stable-abi210-cu130-x86_64-linux: MHA fine, GQA raises
```

```
  .../build/torch-stable-abi210-cu130-x86_64-linux/flash_attn_interface.py:291, in _flash_attn_backward
    ) = flash_attn.bwd(
  .../torch/_ops.py:1279, in __call__
RuntimeError: torch_call_dispatcher( "aten::sum", "IntList_out", stack.data(), (...)) API call failed
  at .../torch-2.13.0/.../torch/include/torch/csrc/stable/ops.h, line 1036
```

| torch | v2 (`cxx11`) | v3 (`stable-abi210`) |
| --- | --- | --- |
| 2.12.1+cu130 | MHA ok, GQA ok | MHA ok, **GQA raises** |
| 2.13.0+cu130 | no CUDA variant | MHA ok, **GQA raises** |

[...]

PR #1125, opened 2026-08-31
flash-attn2: fix dangling pointers in bwd when `num_kv_heads != num_heads`

danieldk:
Also see #1085.


danieldk:
Tested locally. I propose to skip the build and directly going for build-and-merge. The current kernel on `kernels-community` currently already crashes on the patched code path, so we are not introducing worse regressions.

sayakpaul (approved):

sayakpaul:
Since I am already doing the smoke testing sprint for the kernels we maintain in this repo, I think the backward test should also run in CI. I haven't opened a PR for adding smoke tests for FAs yet but will keep it in mind when I do.

PR #1126, opened 2026-08-31
flash-attn2: add smoke tests

danieldk:


flash-attn2/tests/test_flash_attn_ci.py
```diff
@@ -0,0 +1,71 @@
...
+        deterministic=False,
+        mha_type=mha_type,
+        dtype=torch.float16,
+        kvpacked=False,
+        softcap=0.0,
+        device=device,
+    )
+
+
+@pytest.mark.kernels_ci
```
sayakpaul, line 57: Should we have a backward pass test too? Won't add too much time, either.

sayakpaul (approved):

flash-attn2/tests/test_flash_attn_ci.py
danieldk: Note that the tests that are called already verify gradient computation.

danieldk:
(Rebuilding before merge, since we can run tests now.)

PR #1130, opened 2026-09-01
fix json decoding error in the hub comment.

sayakpaul:
Fixes https://github.com/huggingface/kernels-community/actions/runs/33468086649/job/99738296263. 

In our upload logic, we expect the JSON file inside a subfolder: `hub-uploads/some-folder/upload.json. But the corresponding run placed it under `hub-uploads/upload.json`. Hence the mismatch.


drbh (approved):
LGTM!

PR #1131, opened 2026-09-01
feat: implement validation of kernel versions in the tests

sayakpaul:
We use `get_kernel()` in the tests that we run in our CI (and also more generally). 

We should check if the tests use the repo id and the version specified in the `build.toml` of the corresponding kernel being tested. 

I have not run the driver script yet as my plan for this PR is to just add the utilities. We can tackle the checks of the existing kernels in a follow-up PR or through a manually dispatched workflow.


danieldk (approved):
Nice!!! 🔥

PR #1132, opened 2026-09-02
[tests] add FA3 smoke tests

sayakpaul:
* Expose ops so that we can do the testing of backward
* Bump version since we are exposing something new


danieldk:
Building again, because tests should now actually run.

flash-attn3/tests/test_flash_attn.py
```diff
@@ -13,28 +15,26 @@
...
 from .test_util import (
     attention_ref,
     generate_qkv,
     generate_random_padding_mask,
 )
 
+apply_rotary_emb = None
+
+flash_attn3 = kernels.get_kernel("kernels-community/flash-attn3", version=1)
+flash_attn3_ops = importlib.import_module("._ops", package=flash_attn3.__name__)
```
danieldk: This is really flaky. Maybe we should consider exposing `ops`?

danieldk (commented):
Small comment.

flash-attn3/tests/test_flash_attn.py
sayakpaul: Yes 100 percent.

danieldk (approved):
Good to merge if CI is happy.

PR #1135, opened 2026-09-03
CI: set the mode to `pr` on commands that don't upload

danieldk:
The dispatch mode was always set to `release`, which skips kernel tests.


drbh (approved):
thanks for fixing!

PR #1137, opened 2026-09-04
causal-conv1d: add `layers` module for `LayerRepository`

kaixuanliu:
Adds a `layers` submodule to `causal-conv1d`, exposing `causal_conv1d_fn` and `causal_conv1d_update` as `nn.Module` wrappers so the repo can be consumed via `LayerRepository` / `kernelize()`, as transformers will deprecate `FuncRepository` Soon. @danieldk , pls help review, thx!

### DIFF
--- a/causal-conv1d/torch-ext/causal_conv1d/__init__.py
+++ b/causal-conv1d/torch-ext/causal_conv1d/__init__.py
@@ -1,4 +1,12 @@
+from . import layers
 from .causal_conv1d_interface import causal_conv1d_fn, causal_conv1d_update
 from .causal_conv1d_varlen import causal_conv1d_varlen_states
 
-__all__ = ["causal_conv1d_fn", "causal_conv1d_update", "causal_conv1d_varlen_states"]
+__all__ = [
+    # wrappers
+    "layers",
+    # originals
+    "causal_conv1d_fn",
+    "causal_conv1d_update",
+    "causal_conv1d_varlen_states",
+]
--- a/causal-conv1d/build.toml
+++ b/causal-conv1d/build.toml
@@ -1,6 +1,6 @@
 [general]
 name = "causal-conv1d"
-version = 1
+version = 2
 edition = 5
 license = "BSD-3-Clause"
 backends = [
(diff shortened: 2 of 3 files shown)


danieldk (dismissed):

causal-conv1d/torch-ext/causal_conv1d/__init__.py
```diff
@@ -1,4 +1,12 @@
+from . import layers
 from .causal_conv1d_interface import causal_conv1d_fn, causal_conv1d_update
 from .causal_conv1d_varlen import causal_conv1d_varlen_states
 
-__all__ = ["causal_conv1d_fn", "causal_conv1d_update", "causal_conv1d_varlen_states"]
+__all__ = [
+    # wrappers
+    "layers",
```
danieldk, line 7: Since this changes the API, could you bump up the version in `build.toml`?

kaixuanliu: Sure, Done.

vasqu:
It's a bit weird because we already use mamba-ssm and its causal conv1d as layers in transformers no? Not sure what the benefit here is other than additional maintenance 👀

vasqu:
See https://github.com/huggingface/transformers/blob/c676202114bb929ba4377f90fc92c5a8fec72da6/src/transformers/integrations/hub_kernels.py#L173

kaixuanliu:
@vasqu Yes, in transformers we use `mamba-ssm` for cuda. Not sure if the implementation is the same as `causal_conv1D`, if they share the same code, we can move the impl of XPU kernel to this path. @danieldk do you have some comments?

vasqu:
Pretty sure it should be the same implementation (mamba uses it for its fused training path)

danieldk:
Yeah, I think it'd make most sense to use `mamba-ssm`, but I don't think it hurts syncing the layers in this kernel with `mamba-ssm`.

vasqu (approved):
Okay let's do it here as well. Tbh it really doesn't hurt to have it on both sides ig

Would be still great to check what XPU could use on transformers side then 🤗 (you can cc me there then)

danieldk (approved):

PR #1143, opened 2026-09-09
rotary: support partial RoPE in apply_rotary_transformers

piyushumate:
## Description

This PR updates `apply_rotary_transformers` in `rotary` to seamlessly support **partial RoPE** (where `rotary_dim < head_dim`), while maintaining 100% backward compatibility for standard full RoPE.

As discussed with @vasqu in [huggingface/transformers#48622](https://github.com/huggingface/transformers/issues/48622#issuecomment-5603524279):
> *Are we really losing much by just not moving the split and cat into torch. Imo, we can just add another functions similar to https://github.com/huggingface/kernels-community/blob/90423c8b8585e36f9c550cac65fce9a27430c58e/rotary/torch-ext/rotary/layers.py#L7-L41 which then makes a split and cat around it*

Instead of an extra function or separate kernel, `apply_rotary_transformers` now inspects `rotary_dim = min(cos.shape[-1], q.shape[-1])`. It slices and rotates the active `[:rotary_dim]` part in-place, leaving `[rotary_dim:]` pass-through dimensions completely intact.

## Changes
- `rotary/torch-ext/rotary/layers.py`: support `rotary_dim < q.shape[-1]` in `apply_rotary_transformers` module.
- `rotary/torch-ext/rotary/__init__.py`: match the same partial RoPE support in the top-level functional wrapper.
- `rotary/tests/test_rotary.py`: add comprehensive `test_apply_rotary_transformers` across full RoPE and partial RoPE dimensions (`(128, 128)`, `(128, 32)`, `(64, 32)`, `(64, 16)`, `(80, 20)`).

## Verification
- Verified E2E on GPU across all configurations (bfloat16, float32, full RoPE, and partial 25%/50% factors) matching HuggingFace `apply_rotary_pos_emb` mathematical output exactly.

### DIFF
--- a/rotary/tests/test_rotary.py
+++ b/rotary/tests/test_rotary.py
@@ -28,30 +28,8 @@ def apply_rotary_torch(x1: torch.Tensor, x2: torch.Tensor, cos: torch.Tensor, si
     return out1, out2
 
 
-def apply_rotary_torch_wrapper(
-    q, k, cos, sin, conj: bool = False, transformers_style: bool = False, unsqueeze_dim: int = 1
-):
-    """the wrapper for apply_rotary_torch.
-
-    If transformers_style is True, implements the HuggingFace transformers RoPE convention
-    (rotate_half with duplicated frequencies), supporting partial RoPE where cos.shape[-1] <= q.shape[-1].
-    """
-    if transformers_style:
-        cos = cos.unsqueeze(unsqueeze_dim)
-        sin = sin.unsqueeze(unsqueeze_dim)
-        rotary_dim = min(cos.shape[-1], q.shape[-1])
-        q_rot, q_pass = q[..., :rotary_dim], q[..., rotary_dim:]
-        k_rot, k_pass = k[..., :rotary_dim], k[..., rotary_dim:]
-
-        def rotate_half(x):
-            x1 = x[..., : x.shape[-1] // 2]
-            x2 = x[..., x.shape[-1] // 2 :]
-            return torch.cat((-x2, x1), dim=-1)
-
-        q_embed = (q_rot * cos) + (rotate_half(q_rot) * sin)
-        k_embed = (k_rot * cos) + (rotate_half(k_rot) * sin)
-        return torch.cat([q_embed, q_pass], dim=-1), torch.cat([k_embed, k_pass], dim=-1)
-
+def apply_rotary_torch_wrapper(q, k, cos, sin, conj: bool = False):
+    """the wrapper for apply_rotary_torch"""
     rotary_dim = cos.shape[-1]
 
     # apply rotation encoding to Q
@@ -69,16 +47,8 @@ def apply_rotary_torch_wrapper(
     return q_out, k_out
 
 
-def apply_rotary_kernel_wrapper(
-    q, k, cos, sin, conj: bool = False, transformers_style: bool = False, unsqueeze_dim: int = 1, as_module: bool = False
-):
+def apply_rotary_kernel_wrapper(q, k, cos, sin, conj: bool = False):
     """the wrapper for apply_rotary_kernel"""
-    if transformers_style:
-        if as_module:
-            layer = rotary.layers.apply_rotary_transformers()
-            return layer(q, k, cos, sin, unsqueeze_dim=unsqueeze_dim)
-        return rotary.apply_rotary_transformers(q, k, cos, sin, unsqueeze_dim=unsqueeze_dim)
-
     rotary_dim = cos.shape[-1]
 
     # apply rotation encoding to Q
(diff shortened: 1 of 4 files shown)


rotary/tests/test_rotary.py
vasqu, line 84: Could we not extend this test with a new parametrized instead?

rotary/tests/test_rotary.py
```diff
@@ -128,3 +128,50 @@ def test_rotary_equivalence(batch_size, nheads, seqlen, headdim, rotary_dim, qk_
...
         ), "Non-rotated part of K should be unchanged"
+
+
+def rotate_half(x):
+    x1 = x[..., : x.shape[-1] // 2]
+    x2 = x[..., x.shape[-1] // 2 :]
+    return torch.cat((-x2, x1), dim=-1)
+
+
+def apply_rotary_pos_emb_transformers_ref(q, k, cos, sin, unsqueeze_dim=1):
```
vasqu: we can probably wrap similar to apply_rotary_torch_wrapper or include it in there with a kwarg

vasqu (commented):
Just some quick comments, this overall looks correct its mostly design!

vasqu:
Imo we should also bump the version since the behavior does change so it might be confusing if it was used in a different context

rotary/tests/test_rotary.py
```diff
@@ -128,3 +159,57 @@ def test_rotary_equivalence(batch_size, nheads, seqlen, headdim, rotary_dim, qk_
...
+        pytest.param(
+            torch.bfloat16,
+            1e-1,
+            1e-5,
+            marks=pytest.mark.skipif(not supports_bfloat16(), reason="bfloat16 not supported on this GPU"),
+        ),
+    ],
+)
+@pytest.mark.parametrize("as_module", [False, True])
+def test_apply_rotary_transformers(batch_size, nheads, seqlen, headdim, rotary_dim, dtype, atol, rtol, as_module):
```
vasqu: I meant to just expand test_rotary_equivalence instead of a new test 👀

piyushumate: addressed in follow up commit

piyushumate:
@vasqu lemme know what you think

rotary/tests/test_rotary.py
```diff
@@ -65,7 +65,7 @@ def apply_rotary_kernel_wrapper(q, k, cos, sin, conj: bool = False):
 @pytest.mark.parametrize("batch_size", [1, 2])
 @pytest.mark.parametrize("nheads", [8, 16])
 @pytest.mark.parametrize("seqlen", [128, 256])
-@pytest.mark.parametrize("headdim, rotary_dim", [(64, 32), (128, 64), (64, 30)])
+@pytest.mark.parametrize("headdim, rotary_dim", [(64, 32), (128, 64), (64, 30), (128, 32), (64, 16), (80, 20)])
```
vasqu, line 68: Let's add a small comment to highlight that one is the common, and one is the half half case (it's easy to overlook otherwise)

rotary/tests/test_rotary.py
```diff
@@ -128,3 +128,29 @@ def test_rotary_equivalence(batch_size, nheads, seqlen, headdim, rotary_dim, qk_
...
+
+        assert torch.equal(out_q_fn, out_q_mod), "Functional and module outputs should be identical"
+        assert torch.equal(out_k_fn, out_k_mod), "Functional and module outputs should be identical"
+        if rot_dim_total < headdim:
+            assert torch.equal(
+                out_q_fn[..., rot_dim_total:], q_trans[..., rot_dim_total:]
+            ), "Non-rotated part of Q should be unchanged in transformers wrapper"
+            assert torch.equal(
+                out_k_fn[..., rot_dim_total:], k_trans[..., rot_dim_total:]
+            ), "Non-rotated part of K should be unchanged in transformers wrapper"
```
vasqu, line 154: Is this really necessary? Ig you want to sanity check between layer and fn implementations but I don't really feel like we need it?

vasqu (dismissed):
Overall looks good now, just a few smaller nits now!

rotary/tests/test_rotary.py
piyushumate: Done, got rid of it

rotary/tests/test_rotary.py
piyushumate: sounds good, added a comment

piyushumate:
@vasqu addressed the nits. Thanks!

vasqu (dismissed):

rotary/tests/test_rotary.py
vasqu, line 75: this is not on you but could you mark this as ci test? so we can make sure we properly use this for the ci builds

the kernel is fairly old so it didnt apply the right patterns, see e.g. FA2/activation smoke tests and https://huggingface.co/docs/kernels/main/en/builder/writing-kernels#use-getkernel-in-tests

piyushumate:
@vasqu Updated! Replaced the legacy kernel import with kernels.get_kernel("kernels-community/rotary", version=3) and added a lightweight @pytest.mark.kernels_ci smoke test covering both full and partial RoPE, following the FA2/activation pattern.

vasqu:
Sorry Ill be delaying this a bit but not abandoning it! just have some other stuff piled up and would like to just sanity check myself as well

piyushumate:
Lemme know if I can help anyway. Thanks!

PR #1144, opened 2026-09-09
remove auto-dispatch and comment.

sayakpaul:
* Removes the auto-dispatch workflow
* We're now required to use the `kernel-bot` exclusively for testing successful builds
* Adds a workflow so that the GitHub bot can post a helpful comment since we will not run any critical CI automatically


drbh (approved):
removing the dispatch workflow looks good too me!

regarding the new comment bot, its not clear to me how helpful the comment will be - and another approach may be to update the text in the pr template? one of the nice things about removing the dispatch workflow is that we can avoid running an extra action on each pr - however the new comment bot will run each time. I'm happy to include it in this change set, but we might want to revisit

sayakpaul:
> however the new comment bot will run each time. I'm happy to include it in this change set, but we might want to revisit

I think it's easy to miss the PR template, and an added comment on the expected steps might be more helpful. Plus, the workflow is pretty cheap to run.

PR #1147, opened 2026-09-11
sync fa4 and add smoke tests.

sayakpaul:
~Since it's a no-arch kernel, didn't bump the version (need to fix the checker; separate PR).~


flash-attn4/tests/cute/benchmark_block_sparsity.py
```diff
@@ -12,8 +12,14 @@
...
-from flash_attn4.compute_block_sparsity import BlockSparsityKernel
-from flash_attn4.block_sparsity import BlockSparseTensors
+import importlib
+import kernels
+
+flash_attn4 = kernels.get_kernel("kernels-community/flash-attn4", version=0)
+
+importlib.import_module(f"{flash_attn4.__name__}.compute_block_sparsity")
+BlockSparsityKernel = flash_attn4.compute_block_sparsity.BlockSparsityKernel
+BlockSparseTensors = flash_attn4.block_sparsity.BlockSparseTensors
```
danieldk: LLMs have a tendency to do this. Instruct it not to use `importlib` at all, this is a bad pattern. It should use the pattern that we use everywhere else:

```
BlockSparsityKernel = flash_attn4.compute_block_sparsity.BlockSparsityKernel
```

If that doesn't work because `compute_block_sparsity` is not exposed, we should use the special namespace that we discussed in Slack:

https://huggingface.slack.com/archives/C090JN2P8NB/p1788527178843189

I think we haven't settled on the final naming, but we need something like that.

flash-attn4/torch-ext/flash_attn4/interface.py
```diff
@@ -1766,18 +3400,21 @@ def flash_attn_func(
...
     pack_gqa: Optional[bool] = None,
     deterministic: bool = False,
+    score_mod: Optional[Callable] = None,
+    score_mod_bwd: Optional[Callable] = None,
     mask_mod: Optional[Callable] = None,
-    full_block_cnt: Optional[torch.Tensor] = None,
-    full_block_idx: Optional[torch.Tensor] = None,
-    mask_block_cnt: Optional[torch.Tensor] = None,
-    mask_block_idx: Optional[torch.Tensor] = None,
-    block_size: Optional[Tuple[int, int]] = None,
```
danieldk, line 1774: Requires a version bump, since it removes parameters in a public API.

The 'no-arch, doesn't need a version bump' is specific to the case where functions are added or arguments are added that have a default value. If we remove arguments or change the default for an argument, we need. to bump up the version.

flash-attn4/tests/cute/benchmark_block_sparsity.py
danieldk: (Assuming that an LLM is involved, but I personally use an LLM for porting assistance until the porter tool is ready.)

sayakpaul:
@danieldk addressed the feedback. The public check API needs a bit more work which I will do in a separate PR (followed by an update in the instructions to bump version for no-arch kernels).

flash-attn4/torch-ext/flash_attn4/__init__.py
```diff
@@ -1,20 +1,18 @@
...
 
-# Patch cute.compile to optionally dump SASS
-cute.compile = cute_compile_patched
-
+# Nothing else in the kernel imports this module, so it is not reachable as
+# `<kernel>.compute_block_sparsity` unless it is imported here.
+from . import compute_block_sparsity  # noqa: F401
 
 __all__ = [
+    "compute_block_sparsity",
```
danieldk: Do we want to expose this as a public API? Seems like something that would break all the time, requiring API bumps for every small update. I think we should instead make a separate module `_private_for_testing.py`, put

```python
from . import compute_block_sparsity
```

in that module and then this in `__init__.py`:

```python
from . import _private_for_tesing

__all__ = [
    "_private_for_testing",
    ...
]
```

Then the symbols are reachable for tests through `_private_for_testing`, but it is clear that it should not be used by downstream users, avoiding having to bump the version for any change anywhere under `compute_block_sparsity`.

flash-attn4/torch-ext/flash_attn4/benchmark_flash_attention_fp8.py
```diff
@@ -0,0 +1,434 @@
+# Benchmark FP8 attention for FA4 (CuTe-DSL) on SM100.
```
danieldk: Is this file used anywhere inside the kernel? Since it has a `main`, does not seem this should be inside the kernel directory?

danieldk (commented):
Added two more small comments

sayakpaul:
@danieldk I think I have resolved all your comments. Regarding running the tests on Hopper and Blackwell, I did that too. Some tests didn't run because of environment problems and I didn't chase them further (but note that these are just 10ish in number).

danieldk (approved):

PR #1154, opened 2026-09-16
finegrained-fp8: bound the grouped kernels to N and K

### LINKED ISSUE #1153: finegrained-fp8: grouped matmul writes past N and reads past N/K (wrong results on ordinary expert shapes)
## Summary

`matmul_grouped` returns wrong results for a wide range of ordinary expert shapes. The grouped kernels bound their tiles in the M direction only: `store_tile` masks the store with `row_mask` but never with `offs_bn < N`, and the tensor-scale and MX kernels load `b` (and `a`) with no K bound. Whenever a tile extends past `N` or `K`, the kernel reads outside the weight tensor and, because `C` is a contiguous `(S, N)` tensor with `stride_cm == N`, writes the overhanging columns into the *next rows* of the output.

The N overhang is not just an unaligned-shape problem: the autotuner is free to pick `BLOCK_SIZE_N` larger than `N`, so even `N = 128` produces a partial tile with `BLOCK_SIZE_N = 256`. That is why the failures below appear on power-of-two widths too, and why they are intermittent — whether the correct value or the overhang from the previous row lands last in a contended column is unspecified.

The block-scale route is safe: `_w8a8_block_dynamic_fp8_matmul_grouped` asserts `N % block_n == 0` and `K % block_k == 0` and pins `BLOCK_SIZE_N`/`BLOCK_SIZE_K` to those values. The tensor-scale and MX routes have no such assert and take their block sizes from autotune.

## Where

On `main` (`finegrained-fp8/torch-ext/finegrained_fp8/grouped.py`, commit `92b1518`):

- `store_tile` line 138: `tl.store(c_ptrs, c, mask=row_mask[:, None])` — no `offs_bn < N` term, and `offs_bn` comes from `pid_n * BLOCK_SIZE_N + tl.arange(0, BLOCK_SIZE_N)` (line 126) with grid `triton.cdiv(N, META["BLOCK_SIZE_N"])` (lines 559, 684). Shared by all three grouped kernels.
- `w8a8_tensor_dynamic_fp8_matmul_grouped_kernel` lines 302-303: `a`/`b` loaded with no K mask, `b` with no N mask.
- `mxfp_dynamic_matmul_grouped_kernel` lines 408-412: same, plus the unmasked `Bs` load.

## Reproducer

No model, no `transformers`, CUDA only. Every row is a constant power of two, so the FP8/MX quantization of each row is exact and `C[i, j]` must be exactly `K * 2**i` — any deviation is the kernel leaving its logical extent, not a rounding artifact.

```python
import torch

kernel = get_kernel("kernels-community/finegrained-fp8", version=4)
E, ROWS, N, K = 2, 2, 33, 64          # N deliberately not a multiple of BLOCK_SIZE_N
S = E * ROWS
A = (2.0 ** torch.arange(S, dtype=torch.float32)).unsqueeze(1).expand(S, K)
A = A.contiguous().to("cuda", torch.bfloat16)
counts = torch.full((E,), ROWS, device="cuda", dtype=torch.int32)
offsets = counts.cumsum(0).to(torch.int32)

B = torch.ones((E, N, K), device="cuda").to(torch.float8_e4m3fn)          # MXFP8 weights
Bs = torch.ones((E, N, K // 32), device="cuda").to(torch.float8_e8m0fnu)  # UE8M0 scales

C = kernel.matmul_grouped(A, B, Bs, offsets, counts, None)
expected = (A[:, :1].float() * K).expand(S, N)
print(C.float()[:, 0].tolist())        # [64.0, 128.0, 256.0, 0.0]  <- last row clobbered
print(expected[:, 0].tolist())         # [64.0, 128.0, 256.0, 512.0]
torch.testing.assert_close(C.float(), expected, atol=0, rtol=0)
```

Observed on that shape: 31 of 132 elements wrong, all of them row 3 columns 0-30 — exactly the 31-column overhang of row 2's tile (`N = 33`, `BLOCK_SIZE_N = 64`), carrying values accumulated from weight columns that do not exist.

## Scope

Sweep of the same reproducer, 3 repeats per shape, `matmul_grouped` with `output_dtype=None`:

| route | N | K | repeats wrong (stock) | repeats wrong (patched) |
|---|---|---|---|---|
| tensor | 32 | 64 | 0/3 | 0/3 |
| tensor | 33 | 64 | 2/3 | 0/3 |
| tensor | 64 | 64 | 1/3 | 0/3 |
| tensor | 65 | 64 | 3/3 | 0/3 |
| tensor | 128 | 64 | 0/3 | 0/3 |
| tensor | 129 | 64 | 1/3 | 0/3 |
| tensor | 128 | 96 | 3/3 | 0/3 |
| tensor | 128 | 160 | 3/3 | 0/3 |
| mxfp8 | 32 | 64 | 1/3 | 0/3 |
| mxfp8 | 33 | 64 | 1/3 | 0/3 |
| mxfp8 | 64 | 64 | 1/3 | 0/3 |
| mxfp8 | 65 | 64 | 2/3 | 0/3 |
| mxfp8 | 128 | 64 | 1/3 | 0/3 |
| mxfp8 | 129 | 64 | 0/3 | 0/3 |
| mxfp8 | 128 | 160 | 3/3 | 0/3 |
| mxfp4 | 32 | 64 | 1/3 | 0/3 |
| mxfp4 | 33 | 64 | 1/3 | 0/3 |
| mxfp4 | 65 | 64 | 3/3 | 0/3 |
| mxfp4 | 129 | 64 | 1/3 | 0/3 |
| mxfp4 | 128 | 160 | 3/3 | 0/3 |

17 of 24 swept configurations produced wrong output at least once; up to 510 wrong elements in a single call. `K = 96` and `K = 160` fail deterministically on the tensor route (K tail, `BLOCK_SIZE_K` 128); the N-overhang cases are intermittent, as expected from a contended store.

## Environment

- NVIDIA RTX PRO 6000 Blackwell Server Edition, compute capability 12.0, driver CUDA 13.0
- torch 2.14.0+cu130, triton 3.8.0, `kernels` snapshot `8c178950e8710e26a2210e4e909cd02dc16c8715` (`version=4`, `build/torch-cuda`)

`compute-sanitizer --tool memcheck` reports "Device not supported" on this GPU, so I have no sanitizer trace to attach; the evidence above is functional. On a supported device I would expect invalid `__global__` write/read reports in `store_tile` and the `b` loads.

## Fix

What worked here, on top of the `version=4` build:

1. `store_tile`: take `N` and store with `mask=row_mask[:, None] & (offs_bn[None, :] < N)`; pass `N` from all three grouped kernels.
2. Tensor-scale kernel: mask the loop loads with the K tail, `k_mask = offs_k + k * BLOCK_SIZE_K < K`, and add `offs_bn < N` to the `b` load.
3. MX kernel: same K mask for `a`, and bound `b` by clamping its indices rather than masking with `other=` — a single typed fill value cannot serve both packed-uint8 (MXFP4) and fp8 (MXFP8) weights (`other=0` raises `cannot cast int32 to fp8e4nv`). Since `a` is already zero on the K tail and the store drops the N overhang, the clamped reads cannot contribute:

```python
    b_cols = tl.minimum(offs_bn, N - 1)
    b_base = B + expert_id * stride_be + b_cols[None, :] * stride_bn
    ...
        kb = tl.minimum(
            offs_kb + k * (BLOCK_SIZE_K // VALUES_PER_BYTE), K // VALUES_PER_BYTE - 1
        )
        b = tl.load(b_base + kb[:, None] * stride_bk)
```
   and mask the `Bs` load with `(offs_bn[:, None] < N) & sf_mask[None, :]`, `other=127` (UE8M0 1.0).

With those changes all 24 swept configurations are exact (right-hand column above).

Alternatively, if unaligned expert widths are considered out of contract for these routes, the wrappers should assert `N`/`K` alignment the way the block-scale wrapper already does — but that would still leave the `BLOCK_SIZE_N > N` case, which needs the store mask regardless.

Happy to send this as a PR if you prefer.

malaiwah:
Fixes #1153.

`matmul_grouped` returns wrong results for ordinary expert shapes: the grouped kernels bound their tiles in the M direction only, so a tile runs past `N` whenever `BLOCK_SIZE_N` does not divide it — or simply exceeds it, which autotune is free to pick — and past `K` whenever `BLOCK_SIZE_K` does not divide it. Since `C` is a contiguous `(S, N)` tensor, the unmasked store writes the overhang into the *following rows*: the corruption lands in another token's output rather than in allocator slack.

## What changed

- `store_tile` takes `N` and masks the store with `offs_bn < N`. All three grouped kernels pass it.
- The tensor-scale kernel masks its `a` / `b` loads with the K tail, and `b` with `offs_bn < N`.
- The MX kernel masks `a` with the K tail and bounds `b` by *clamping* its indices rather than masking with `other=`: one typed fill value cannot serve both packed-uint8 (MXFP4) and fp8 (MXFP8) weights (`other=0` raises `cannot cast int32 to fp8e4nv`). Since `a` is already zero on the K tail and the store drops the N overhang, the clamped reads cannot contribute. Its `Bs` load is masked with `other=127` (UE8M0 1.0).

The block-scale route already asserts `N % block_n == 0` and `K % block_k == 0` and pins its block sizes, so it is unaffected — the change there is only the extra `N` argument.

## Test

`finegrained-fp8/tests/test_grouped_tails.py` drives `matmul_grouped` over the three routes and six `(N, K)` shapes. Rows carry distinct powers of two, so each product is exact and the expected value is just `K * 2**row`: a failure means the kernel left its extent, never rounding. Each shape runs three times, because the overhang and the legitimate write target the same addresses — a single clean call proves nothing about which store retires last.

On an RTX PRO 6000 Blackwell (SM120), torch 2.14.0+cu130 / triton 3.8.0:

| | before | after |
|---|---|---|
| `tests/test_grouped_tails.py` | 13 failed, 5 passed | 18 passed |

A failing case returns `[64.0, 128.0, 256.0, 0.0]` in column 0 instead of `[64.0, 128.0, 256.0, 512.0]` — the last row's low columns overwritten by the previous row's overhang, which itself accumulated weight columns that do not exist.

The existing `tests/test_moe.py` problem set uses `N` of 256/512/1024/1536/2048/4096 throughout, which is why this never showed up there.


IlyasMoutawwakil:
Real bug and the fix looks right.

Two things I'd check before it lands. Masked loads with other=0.0 in the K loop block WS lowering in these kernels, you get 'arith.constant' op does not have expected attribute ttg.partition and the pipeliner pass fails; finegrained-kernels had to fence its mx-grouped affine arm off for exactly that. If the WS configs stop compiling the tuner just crowns something else and nothing fails visibly. Separately, every aligned shape now pays an inner-loop mask it didn't before, before/after on one or two of the shapes already in the problem set would say whether it's worth predicating on a constexpr "tail exists" flag.

Two smaller ones. I'd leave the block-scale route asserting rather than masking, with a comment saying why: a partial N tile has no scale for its incomplete 128-block, so a store mask hides that instead of fixing it. And I'd pin the tiles in the test, 320 is divisible by 64, so if the tuner crowns a dividing BLOCK_SIZE_N the tail path never runs.

malaiwah:
Thanks for the pointers!

malaiwah:
Thanks for the careful read — you were right on everything. Pushed `c7ec61f2` (one squashed commit on top of the original two).

**Aligned shapes / inner-loop mask.** `matmul_grouped` on aligned problem-set shapes, RTX PRO 6000 Blackwell (SM120), torch 2.14.0+cu130 / triton 3.8.0, `do_bench` median µs, two independent rounds (agree within ~1%):

| route | E / S / N / K | main | first push (masks in loop) | now |
|---|---|---|---|---|
| tensor | 8 / 4096 / 2048 / 2048 | 109.6 | 106.5 | 107.2 |
| tensor | 8 / 4096 / 1536 / 4096 | 173.1 | 168.7 | 171.2 |
| tensor | 32 / 8192 / 512 / 2048 | 96.2 | 97.4 | 96.9 |
| mxfp8 | 8 / 4096 / 2048 / 2048 | 159.1 | **367.8** | 157.8 |
| mxfp8 | 8 / 4096 / 1536 / 4096 | 273.9 | **638.0** | 270.3 |
| mxfp8 | 32 / 8192 / 512 / 2048 | 119.8 | **220.7** | 119.5 |
| mxfp4 | 8 / 4096 / 2048 / 2048 | 268.6 | **447.6** | 241.7 |
| mxfp4 | 8 / 4096 / 1536 / 4096 | 507.0 | **751.1** | 440.5 |
| mxfp4 | 32 / 8192 / 512 / 2048 | 190.4 | **297.5** | 170.8 |

So the first version cost 1.5–2.3× on every MX shape (the masked `Bs` load with `other=127` plus the per-step clamp, I assume — the tensor route didn't care). The new version is at parity with `main` on tensor/mxfp8 and a bit faster on mxfp4, which I'd attribute to the tuner finding a different config now that the loop body is clean rather than anything clever.

What changed to get there: the N tail is bounded by wrapping the weight/scale column index (`offs_bn % N`) once before the loop — inert when `BLOCK_SIZE_N` divides `N`, and the wrapped columns only ever land in accumulator columns the store mask drops. The K tail is peeled into a scalar-gated block after the loop, so the loop body is the stock code again and only carries the `row_mask` it always had. I went with peeling rather than a constexpr flag, but an inner `@triton.heuristics` deriving `EVEN_K` from the crowned config would work too if you'd rather have that shape — happy to switch.

**WS lowering.** Fair warning, and it's why the K-tail masks are out of the loop now. For the record these kernels use plain `range()` and never opt into `tl.range(warp_specialize=...)`, so nothing was silently dropping today; I didn't want to leave the trap for whoever adds it.

**Block-scale route.** Kept asserting, with your reasoning in the comment.

**Pinning.** The test now pins `BLOCK_SIZE_N=64 / BLOCK_SIZE_K=128` through the tuner's `configs`, and I added `(33, 128)` (N tail, no K tail) and a `dot`-mode pin at its real `BK=32`. That made detection deterministic: the same test file against `main`'s kernels fails 25/25 on SM120, where my unpinned run was getting 13/18.

**One thing I broke and found while doing this:** the first push changed `store_tile`'s signature but not the two callsites in `fused_grouped.py`, so every fused MoE path was failing to compile. Ouch, my bad, sorry. Fixed and all good now. (Well it uncovered an distinct existing issue I will file separately)

malaiwah:
For reference, the same bug class happens one level up as well, see https://github.com/huggingface/kernels-community/pull/1159

PR #1156, opened 2026-09-16
[MPS] Fix metal flash sdpa encoder handoff

### LINKED ISSUE #1155: metal-flash-sdpa crashes after a PyTorch MPS operation
I discovered this while investigating transformers `cu_seq_lens_k.clone()` workaround here:
https://github.com/huggingface/transformers/blob/51b330bb5c30ed3f6ab2632961b3f88c45293686/src/transformers/modeling_flash_attention_utils.py#L779-L782
which (the op itself not clone) opens a second compute encoder while torch's encoder is active on the same command buffer.  Reason cloning fixes this issue is because cloning uses a blit encoder, which ends the compute encoder first so bug is masked.

Reproducer:
```
import torch
import torch.nn.functional as F

kernel = get_kernel("kernels-community/metal-flash-sdpa", version=1)

q, k, v = [torch.randn(20, 4, 64, device="mps") for _ in range(3)]
lengths = torch.tensor([8, 12], dtype=torch.int32, device="mps")
out = torch.empty_like(q)
torch.mps.synchronize()

cu = F.pad(lengths.cumsum(0, dtype=torch.int32), (1, 0))
q = torch.sin(q)  # leaves torchs compute encoder open
kernel.flash_attention_varlen(
    out, q, k, v, cu, cu, 12, 12, False, 64**-0.5, 1.0
)
torch.mps.synchronize()
```

gives me:
```
/Users/salia/Desktop/pytorch-2/.torch_old/lib/python3.14/site-packages/torch/_subclasses/functional_tensor.py:368: UserWarning: Failed to initialize NumPy: No module named 'numpy' (Triggered internally at /Users/runner/work/pytorch/pytorch/torch/csrc/utils/tensor_numpy.cpp:84.)
  cpu = _conversion_method_template(device=torch.device("cpu"))
Fetching 6 files: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████| 6/6 [00:00<00:00, 2458.08it/s]
Download complete: :                                                                                                                     |  0.00B            
-[AGXG17XFamilyCommandBuffer tryCoalescingPreviousComputeCommandEncoderWithConfig:nextEncoderClass:]:1168: failed assertion `A command encoder is already encoding to this command buffer'
zsh: abort      python temp.py
```

I have a change locally which reuses PyTorch's encoder which should fix this issue, if we agree on pursuing this fix. I think this is problematic for other attention implementations as well but I haven't looked too deep into it

Isalia20:
Fixes https://github.com/huggingface/kernels-community/issues/1155

Bug and motivation is explained in the above issue but TLDR is that instead of creating a new compute encoder on the command buffer created by Torch (which fails when another encoder is already active), we reuse the compute encoder of torch directly. The blit path ends compute coalescing before opening its own encoder.

## Checklist

- [x] This PR is linked to an issue that was discussed and approved
- [x] I have tested these changes locally
- [x] New/changed functionality has test coverage
- LLM disclosure:
  - [x] I did not use an LLM to create this PR.
  - [x] I used and LLM for assistance while creating this PR.
  - [x] This PR was mostly or completely generated by an LLM.


EricLBuehler (approved):
Looks great! 🚀

It looks like CI is blocked by an existing issue that we didn't run into before, perhaps you can fix this?

danieldk (approved):

danieldk:
/kernel-bot merge-and-upload metal-flash-sdpa paged-attention

PR #1160, opened 2026-09-17
finegrained-kernels: Optimize for Intel XPU

kaixuanliu:
Make the kernels, tests and benchmark run on Intel XPU (torch-xpu / Triton XPU):

- compat: omit the unsupported `launch_pdl=` launch kwarg on non-CUDA backends (`pdl_launch_kwargs()`), derive `sm_count()` / `persistent_program_count()` / `sm_shared_memory_limit()` from XPU device properties, and narrow the autotune grid to block sizes that compile on XPU.
- grouped: use `persistent_program_count()` for the persistent launches and fall back to affine (unswizzled) MX scales on XPU, with a cached unswizzle helper in swizzle.py.
- pruners: cap the GATE tile to BLOCK_SIZE_N<=32 on XPU (measured winner).
- backward: default WARP_SPEC to False (unsupported on XPU).
- tests/bench: drop the CUDA hardcoding in favour of `torch.accelerator`, and build MoE weights per expert so the suite fits in 32 GiB of device memory.

@IlyasMoutawwakil Pls help review, thx!


kaixuanliu:
Latest speedup benchmark result:
xpu_speedup

finegrained-kernels/torch-ext/finegrained_kernels/grouped.py
```diff
@@ -1720,6 +1720,22 @@ def mx_dynamic_matmul_grouped(
...
+    if (
+        swizzled_scales
+        and prefer_affine_mx_scales()
+        and n_rows % 128 == 0
+        and (As is None or As.ndim != 5)
+    ):
+        Bs = unswizzle_mx_scales_cached(
+            Bs, n_rows, K // scale_group, num_experts=num_experts
+        )
+        swizzled_scales = False
```
IlyasMoutawwakil: better not have them swizzeled at all for xpu then, this hits perf

finegrained-kernels/torch-ext/finegrained_kernels/batched.py
```diff
@@ -1257,7 +1259,7 @@ def w8a8_block_static_fp8_matmul_batched(
             SIMULATE_UNFUSED=simulate_unfused,
             INTERMEDIATE_DTYPE=tl_dtype(output_dtype),
             PDL=decode_pdl(),
```
IlyasMoutawwakil: this should be in the pdl kwargs too

finegrained-kernels/tests/utils.py
```diff
@@ -143,35 +141,53 @@ def make_weights(
...
+    for _ in range(E):
+        W = torch.randn(1, out_features, in_features, dtype=torch.float32, device=device)
+        W = torch.nn.functional.pad(W, [0, pad_k, 0, pad_n])
+        R = W.reshape(1, Np // block_n, block_n, Kp // block_k, block_k)
+        del W
+        max_abs = R.abs().amax(dim=(-3, -1))
+        safe = torch.where(max_abs > 0, max_abs, torch.ones_like(max_abs))
+        inv_e = (safe / qmax).to(torch.float32)
+
+        exp_e = None
```
IlyasMoutawwakil: ouch, this is slower on b200 where we test these kernels

kaixuanliu:
For the performance fallback w/ BF16 prefill, after investigation, it shows the loss is purely GEMM throughput. On XPU, Triton only emits the fast 2D block-load (the one that can feed DPAS at rate) when a tile is addressed via tl.make_block_ptr, but finegrained-kernels uses plain pointer arithmetic everywhere — zero make_block_ptr in the whole package. That costs 2.9x of GEMM throughput at the real per-expert prefill shape: 49.4 vs 143.0 TFLOP/s measured, where 143.0 is 97% of the oneDNN ceiling. 

And unrelated: the 0.48x compile figure is an artifact — finegrained-kernels compile fails on every MoE row (XPUDriver.get_active_torch_device should not be traced), so that column is our-eager vs their-compile.

finegrained-kernels/torch-ext/finegrained_kernels/grouped.py
kaixuanliu: Well, agree this. no-op is enough.

kaixuanliu:
Add an XPU-only block-pointer arm to the full-precision grouped MoE GEMM that describes the (E, N, K) weight tile in its native K-contiguous orientation and transposes in registers, letting Xe emit 2D block loads — ~2x BF16 MoE prefill, bit-identical. Here is the updated benchmark result for BF16 MOE:
| problem | regime | speedup |
|---|---|---|
| small/DSV4-shaped | prefill | 1.34x |
| openai/GPT-OSS-120B | prefill | 1.03x |
| small/DSV4-shaped | decode | 2.76x |
| deepseek-ai/DSV4 | decode | 3.12x |
| openai/GPT-OSS-120B | decode | 2.93x |

finegrained-kernels/torch-ext/finegrained_kernels/batched.py
kaixuanliu: Fixed.

finegrained-kernels/torch-ext/finegrained_kernels/grouped.py
kaixuanliu: Fixed.

finegrained-kernels/bench/bench_moe.py
```diff
@@ -344,6 +350,19 @@ def _check_cuda_home():
...
+    "small/DeepSeek-V4-shaped FP8 block-dyn W8A8 ue8m0 (E32 H2048 I1024 top6)": dict(
+        E=32, H=2048, I=1024, top_k=6, weights="fp8_128x128_ue8m0", activation_format=None,
+        baselines=("finegrained-fp8", "deepgemm"), fp8_block=[128, 128], block_size=(128, 128),
+        act="silu", swiglu_alpha=None, swiglu_limit=None,
+    ),
+    "small/MiniMax-M3-shaped MXFP8 (E32 H2048 I1024 top4)": dict(
+        E=32, H=2048, I=1024, top_k=4, weights="mxfp8", activation_format=None,
+        baselines=("finegrained-fp8",), fp8_block=None, block_size=None,
+        act="silu", swiglu_alpha=1.702, swiglu_limit=7.0,
+    ),
```
IlyasMoutawwakil, line 365: this only adds smaller models ? isnt the idea to replace big ones with the smaller ones on low vram devices ?

finegrained-kernels/bench/bench_moe.py
```diff
@@ -334,7 +334,13 @@ def _check_cuda_home():
...
     _tfmx.triton_kernels_hub = triton_kernels_hub
 
-DEV = "cuda"
+DEV = torch.accelerator.current_accelerator().type if hasattr(torch, "accelerator") else "cuda"
+# Only the device-visibility knob for the sharded runner is backend-specific; every other
+# accelerator call goes through torch.accelerator.
+DEV_MASK_ENV = {"cuda": "CUDA_VISIBLE_DEVICES", "xpu": "ZE_AFFINITY_MASK"}.get(DEV)
+if DEV_MASK_ENV is None:
+    raise RuntimeError(f"unsupported accelerator {DEV!r}; this benchmark needs cuda or xpu")
+ACCEL = torch.get_device_module(DEV)
```
IlyasMoutawwakil, line 343: use full name ACCELERATOR

finegrained-kernels/bench/bench_moe.py
```diff
@@ -334,7 +334,13 @@ def _check_cuda_home():
     triton_kernels_hub = get_kernel("kernels-community/gpt-oss-triton-kernels", version=1)
     _tfmx.triton_kernels_hub = triton_kernels_hub
 
-DEV = "cuda"
+DEV = torch.accelerator.current_accelerator().type if hasattr(torch, "accelerator") else "cuda"
+# Only the device-visibility knob for the sharded runner is backend-specific; every other
+# accelerator call goes through torch.accelerator.
+DEV_MASK_ENV = {"cuda": "CUDA_VISIBLE_DEVICES", "xpu": "ZE_AFFINITY_MASK"}.get(DEV)
+if DEV_MASK_ENV is None:
+    raise RuntimeError(f"unsupported accelerator {DEV!r}; this benchmark needs cuda or xpu")
```
IlyasMoutawwakil, line 342: the check for the accelerator can be done on DEVICE before the mask

finegrained-kernels/tests/utils.py
kaixuanliu: Well, reverted. Will do the tests on XPU offline

finegrained-kernels/tests/utils.py
```diff
@@ -60,11 +58,11 @@
...
-    if TEST_DEVICE == "cuda":
-        return torch.cuda
-    if TEST_DEVICE == "xpu":
-        return torch.xpu
-    raise RuntimeError("No supported accelerator available")
+    """The backend module (torch.cuda / torch.xpu) for the active accelerator, for the few
+    calls ``torch.accelerator`` does not expose (e.g. ``get_device_name``)."""
+    if TEST_DEVICE is None:
+        raise RuntimeError("No supported accelerator available")
+    return torch.get_device_module(TEST_DEVICE)
```
IlyasMoutawwakil, line 65: isnt torch.accelerator the same as accelerator_module ?

finegrained-kernels/torch-ext/finegrained_kernels/backward.py
```diff
@@ -215,7 +215,7 @@ def dgrad_matmul_2d_kernel(
     BLOCK_SIZE_M: tl.constexpr,
     BLOCK_SIZE_N: tl.constexpr,
     BLOCK_SIZE_K: tl.constexpr,
-    WARP_SPEC: tl.constexpr,
+    WARP_SPEC: tl.constexpr = False,  # tuner axis on CUDA only; off-CUDA grids never emit it
```
IlyasMoutawwakil, line 218: better make the config generator return only False on this field, i prefer that over a default bcz a default can easily silcence a missing optim knob on hw that supports it

finegrained-kernels/torch-ext/finegrained_kernels/swizzle.py
```diff
@@ -158,6 +158,10 @@ def swizzle_mx_scales(
     ``(G, row_blocks, ceil(cols/4), 2, 256)`` artifact the ops read — ``G`` the expert count (1 for
     a matrix), so an expert stack shards/gathers on its leading dim. Bit-identical to CUTLASS's
     packer (verified)."""
+
+    # XPU returns ``scale`` untouched, so the ops take their row-major (affine) scale arms.
+    if get_active_device_type() == "xpu":
+        return scale
```
IlyasMoutawwakil, line 164: the function swizzle_scales should swizzle the scales, it's better it not be called on xpu rather than it returning the scale as is

finegrained-kernels/bench/bench_moe.py
kaixuanliu: Fixed

finegrained-kernels/bench/bench_moe.py
kaixuanliu: Fixed

finegrained-kernels/tests/utils.py
kaixuanliu: Fixed

finegrained-kernels/torch-ext/finegrained_kernels/backward.py
kaixuanliu: Fixed

finegrained-kernels/bench/bench_moe.py
```diff
@@ -343,7 +351,12 @@ def _check_cuda_home():
 # support first, finegrained-kernels-only (GPT-OSS, GLM-NVFP4) last.
 CANONICAL_MODEL_ORDER = ["DeepSeek-V4", "DeepSeek-V3", "MiniMax-M3", "GPT-OSS-120B", "GLM-5.2"]
 
-MOE_PROBLEMS = {
+# Opt-in for low-VRAM parts: the same recipes and baselines at /8 experts, /2 dims, run
+# INSTEAD OF the full roster rather than alongside it. Off by default.
+SMALL_GEOMETRY = os.environ.get("FGK_BENCH_SMALL_GEOMETRY", "0") == "1"
```
IlyasMoutawwakil: just check the device's memory vram/hbm

finegrained-kernels/bench/bench_moe.py
kaixuanliu: Fixed

finegrained-kernels/torch-ext/finegrained_kernels/swizzle.py
kaixuanliu: Fixed

finegrained-kernels/bench/bench_moe.py
kaixuanliu: Done.

finegrained-kernels/tests/utils.py
```diff
@@ -47,6 +45,10 @@
 # SM90 (Hopper, e.g. H100) is the only architecture the benchmark baselines
 # are calibrated against — every other SM has its own latency profile.
 IS_SM90 = TEST_DEVICE == "cuda" and torch.cuda.get_device_capability() == (9, 0)
+# The SWIZZLE_32_4_4 scale layout exists to feed tcgen05, so the ops only grow a
+# swizzled arm on CUDA. Elsewhere the scales stay row-major (affine) and the
+# pre-swizzled variants of each problem are skipped rather than run unswizzled.
+SUPPORTS_SWIZZLED_SCALES = TEST_DEVICE == "cuda"
```
IlyasMoutawwakil: maybe use the is_sm100 util

finegrained-kernels/torch-ext/finegrained_kernels/grouped.py
```diff
@@ -1219,18 +1219,51 @@ def full_precision_matmul_grouped_kernel(
...
+                    a_ptrs, ADescriptor, m_start, k * BLOCK_SIZE_K, row_mask, in_row,
+                    A_MEMORY_MODE, GatherIdx is not None,
+                )
+                w, _ws = load_weight_plain(
+                    b_ptrs, BDescriptor, row0, n_off, k * BLOCK_SIZE_K,
+                    GATE, True, B_MEMORY_MODE, False, BLOCK_SIZE_N, BLOCK_SIZE_K,
+                )
+                acc = acc + fp8_dot(a, w, False, BLOCK_SIZE_K)
+                a_ptrs += BLOCK_SIZE_K * stride_a_k
+                b_ptrs += BLOCK_SIZE_K * stride_b_k
```
IlyasMoutawwakil: can you please integrate this into the helpers we use ? load_act_plain, maybe a helper that take the original descriptor + the memory mode and returns either the original host descriptor or creates a new device one. that way we keep the layout of the gemm simple (that's actually the whole point of the helpers, the gemm routines layout stays simple, the implementation specifities are managed seperately)

PR #1163, opened 2026-09-18
feat: error early when a version branch doesn't exist in external repos.

sayakpaul:
Currently, we have a couple of kernels for which we open PRs outside of the kernels-community Hub org. 

A kernel build usually has two branches to target:

* `main`
* version branch

In some cases, we bump the version of a kernel and that means the build is uploaded in a new branch. For uploads to the kernels-community Hub org, this is fine because we have write access to create the requested branches. However, for external repositories, we cannot do this. PRs are valid for existing branches only. 

We should error out early in this case to skip the (often extensive) build process for commands like `release` and `merge-and-upload`. I think this is the right design because:

* Usually, on PRs, we run `kernels-bot build ...` first to validate if things are looking good. So, when we ultimately run `merge-and-upload`, if we violate the branch conditions for PRs, it makes sense to stop and error out early. 
* Same applies for the "release" command as well.


drbh (approved):
lgtm

PR #1164, opened 2026-09-18
Version-gate the warp-spec fence and fully extend and test static quant (mistal4)

IlyasMoutawwakil:
Two changes to the block-scale FP8 and MX grouped/batched paths.

WARP-SPEC FENCE. `warp_spec_memory_mode_pruner` dropped configs the WS passes cannot lower, from a table charted on one compiler. Re-charting on B200 against BOTH toolchains, by filtering each kernel's own tuner configs with BM and num_warps as axes, shows the two are near mirror images:

    operands          triton 3.8.0             triton 3.7.1
    pointer/pointer   every warp count and BM  mostly FAILS
    desc/desc         ONLY w4 AND BM == 128    most cells lower; w2 never
    mixed             same as desc/desc        same as desc/desc

Cell for cell over three kernels, 3.8 rescues NO descriptor cell 3.7.1 lowered (0 of 96) and 3.7.1 rescues NO pointer cell 3.8 lowered (0 of 80): descriptor+WS REGRESSED in 3.8, pointer+WS was FIXED. So each version gets the rule its own sweep measured. The pre-3.8 pointer traps stay because 3.7.1 reproduced them -- they were a correct chart, not the mis-chart the descriptor rows turned out to be.

PER-EXPERT STATIC ACTIVATION SCALES. A MoE calibrates each expert separately, so `activation_scheme="static"` gives one scale per expert, while these kernels took one per matmul. The scale now rides `stride_as_e` -- 0 means every expert reads the one calibrated value, 1 gives each its own -- so there is no second specialization and the scalar path is unchanged (verified bit-identical).

Per expert the activations cannot be pre-quantized on the host: top-k routes one row to several experts whose scales differ, so it has no single pre-quantized form. The wrapper hands A raw and the kernel quantizes in register against the tile's own expert, which also removes the host pass the scalar path pays. `load_act_static` grows the arm that makes this possible: a raw tile behind a gather, through the same load the fp8 arm uses.

Tested over {scalar, per-expert} x {expert-sorted, gathered} against a reference that quantizes each expert's rows by its own scale. Reverting the expert index fails both per-expert cases and neither scalar one. The gap that hid this: the suite covered `static=True` on all three ops, but `make_static_activation_scale` only ever built a per-tensor scalar, so A always arrived pre-quantized and the raw arm was never reached.

## Important: Read before submitting

> **New contributors:** Please open or comment on an issue **before** submitting a
> PR to discuss the change you'd like to make. This helps us align on approach and
> avoids wasted effort on changes we may not be able to merge. Please only create
> a PR once one of the project maintainers agrees on your outlined approach.
>
> PRs of contributors who are not vouched for are automatically closed. Regular
> contributors are added to the vouch list.
>
> The kernels-community repository is for:
>
> 1. Kernels developed by Hugging Face and partners.
> 2. Kernels developed by third parties, that have not been 'kernelized' yet, but
>    are used by Hugging Face projects such as diffusers and transformers.
>
> We cannot accept PRs for the following:
>
> - New kernels that are not in categories (1) and (2).
> - Changes to kernels in category (2), these should be submitted upstream.
>
> For, LLM-generated changes, we prefer that you write your prompt in an issue
> over a PR with LLM-generated changes.

## Related issue

Closes #

## Changes

-

## Testing

-

## Checklist

- LLM disclosure:


finegrained-kernels/torch-ext/finegrained_kernels/compat.py
```diff
@@ -40,6 +40,14 @@ def decode_pdl() -> bool:
     no fusion). Deployment decode runs eager launches under cudagraphs, where PDL applies."""
     return DECODE_PDL and not torch.compiler.is_compiling()
 
+# The scaled_grouped_mm scaling enums the torch MoE baseline hands its scale operands. Older
+# torch has neither them nor the op, and only that one baseline reads them, so a missing pair is
+# ``None`` here and the baseline refuses on it rather than the whole package failing to import.
+try:
+    from torch.nn.functional import ScalingType, SwizzleType
+except ImportError:
+    ScalingType = SwizzleType = None
```
vasqu, line 49: when was it introduced? Maybe a proper check on the torch version than this but no biggie

vasqu (dismissed):
Imo looks good, I'm just wondering about the per expert calibration where is this needed or what kind of model does it?

I mean the shared is essentially the old style static

IlyasMoutawwakil:
"I'm just wondering about the per expert calibration where is this needed or what kind of model does it?"
yeh shared is just basically if the quantization was done on the packed gate_up/down experts but i think it's never the case, you wanna calibrate for each expert separately. i will drop that distinction.

IlyasMoutawwakil:
it's basically one "shared" scale for dense and per-expert for moe, so crossing them is unnecessary

finegrained-kernels/torch-ext/finegrained_kernels/compat.py
IlyasMoutawwakil: it was used in the pure torch path, just imported in the function body

vasqu:
Yea that makes much more sense

finegrained-kernels/bench/bench_moe.py
```diff
@@ -128,15 +129,27 @@
...
+    try:
+        _kernels_mod = importlib.import_module(_mod)
+    except ModuleNotFoundError:
+        continue
+    if hasattr(_kernels_mod, "_check_trust_remote_code"):
+        _kernels_mod._check_trust_remote_code = lambda *a, **k: None
+        _patched.append(_mod)
+if not _patched:
+    print(f"[bench] no _check_trust_remote_code in kernels {kernels.__version__}: "
+          "baseline loads may hit the rate-limited publisher-trust API")
```
vasqu: I know its just a bench scrip but tbh we can just do it properly? esp since the kernel should be staged to be used instead of fully locally

finegrained-kernels/torch-ext/finegrained_kernels/grouped.py
```diff
@@ -2234,12 +2252,18 @@ def matmul_grouped(
...
-    if As is not None and As.numel() == 1:
-        # static (per-tensor calibrated) activation quant: a per-tensor scalar As for block-scale FP8
-        # weights — the caller hands raw A, the op quantizes it against the scalar (As IS the scale).
-        assert Bs is not None and not is_mx(B, Bs) and weight_block_size(B, Bs) is not None, (
-            "a per-tensor scalar As (static activation scale) needs block-scale FP8 weights"
+    # static (calibrated) activation quant: the caller hands raw A and As IS the scale — one value
+    # for the whole matmul, or one per expert for a MoE, which calibrates each expert separately.
+    # A pre-quantized A carries per-token scales instead, and per-block ones are 2-D; both take the
+    # dynamic arms below. Block-scale weights have a dedicated static kernel, per-tensor ones read
+    # the same As on the tensor-wide arm.
```
vasqu: the comments are repeating a bit :D (over the file)

finegrained-kernels/torch-ext/finegrained_kernels/batched.py
```diff
@@ -94,30 +94,6 @@ def expert_setup(
     return batch_id, pid_n, expert_id, A, B, C, Bs, in_row, out_row
 
 
-@triton.jit
-def store_row(
```
vasqu, line 98: so missed one epilogue or something -> unneeded?

vasqu (dismissed):
Just nits

finegrained-kernels/bench/bench_moe.py
IlyasMoutawwakil: this is because the kernels changed apis between PRs and i just want the bench to work, what's properly exactly ?

finegrained-kernels/torch-ext/finegrained_kernels/batched.py
IlyasMoutawwakil: this store was mainly used by the static batched when it didn't suport fused epilogues, so yes notneeded now

finegrained-kernels/bench/bench_moe.py
vasqu: Not trying to side effect trust remote code :p yea not super important it just looks a bit ugly

IlyasMoutawwakil: i will just force latest kernels then

finegrained-kernels/torch-ext/finegrained_kernels/grouped.py
IlyasMoutawwakil: it's repeated in different files batched vs grouped

finegrained-kernels/tests/test_autotuner.py
```diff
@@ -266,12 +299,18 @@ def test_cross_process_determinism_block_dynamic_grouped():
...
     shape key must skip the doomed compile and report the failure as memoized. Bench
     -stage errors are excluded by design (a sticky-context cascade must never fence
-    healthy configs)."""
+    healthy configs).
+
+    The suite pins ``FINEGRAINED_AUTOTUNE_MAX_FAILURES=0`` so a config the tuner cannot compile
+    fails loudly instead of being forgiven (conftest). This test IS that machinery and compiles
+    a deliberately doomed config, so it restores the deployment default for its own tune —
+    otherwise the first reject aborts the tune before there is anything to memoize."""
+    monkeypatch.setenv("FINEGRAINED_AUTOTUNE_MAX_FAILURES", "3")
```
vasqu: this is weird, can we really only control it with the env flag and not make it passable as arg?

finegrained-kernels/torch-ext/finegrained_kernels/loading/tiles.py
```diff
@@ -268,15 +268,18 @@ def load_act_static(
...
-    else:  # raw bf16/fp16 (inline arm, M<threshold, pointer-only) — quantize vs the static scale
-        a = (tl.load(a_ptrs).to(tl.float32) / a_s_static).to(tl.float8e4nv)
+    """static activation tile, through the same tile load in every memory mode: a pre-quantized
+    fp8 A is the rows-major MMA lhs as-is, raw bf16/fp16 is quantized in register against
+    ``a_s_static``. The raw arm is what a PER-EXPERT scale needs: one row can be routed to
+    several experts whose calibrated scales differ, so it has no single pre-quantized form. Its
+    own load would have to repeat the tile loader's masking and descriptor arms — a bare
+    pointer load reads past the tensor on a grouped tail tile. ``a_s`` = the values (the scale
+    is folded post-loop)."""
+    a = load_grouped_act_tile(
```
vasqu, line 278: so we are guaranteed to work with any dtype here, i.e. the guard before is no longer necessary?

finegrained-kernels/torch-ext/finegrained_kernels/quant.py
```diff
@@ -930,6 +930,102 @@ def _fp8_act_quant_kernel(
 @compile_time_only_triton_op(
     add_op_namespace_prefix("fp8_act_quant_tensor_wide"), mutates_args=(), opaque=True
 )
```
vasqu: decorator still correct?

finegrained-kernels/torch-ext/finegrained_kernels/quant.py
```diff
@@ -930,6 +930,102 @@ def _fp8_act_quant_kernel(
...
+    A per-expert scale otherwise hands the kernel raw rows to quantize per tile, which reads
+    ``A`` at bf16 width through every N-tile and holds the tile at ``RAW_ACT_MAX_BN``. Laying
+    the rows out once buys back both, at the rounding the in-register arm produces, because a
+    row's expert is known once the rows are expert-sorted — ``gather_idx`` orders them, or the
+    intermediate already is. ``As`` is untouched: the kernel dequantizes per expert whichever
+    form ``A`` arrives in.
+
+    Returns ``(A, gather_idx)``, ``gather_idx`` ``None`` once the rows are laid out.
+    """
+    return quantize_rows_static(A, As, gather_idx, expert_start, num_experts), None
```
vasqu: why are we even returning the none at this point?

finegrained-kernels/torch-ext/finegrained_kernels/quant.py
vasqu: kind of weird if this still worked -> we may have quite a few unnecessary things in the code base

finegrained-kernels/torch-ext/finegrained_kernels/batched.py
```diff
@@ -493,14 +488,21 @@ def w8a8_tensor_dynamic_fp8_matmul_batched_kernel(
...
-    )
+    # per-tensor scales fold on the raw accumulator; `gemm_epilogue` owns the finalize.
+    # Finalizing here too was invisible under no-swap (a pass-through) and collapsed the
+    # swapped tile twice, so every SWAP_AB config failed to compile and the tuner quietly
+    # dropped the whole swap arm at this shape.
+    accumulator = accumulator * a_s * b_s
     if PDL:
         gdc_launch_dependents()
-    store_row(C, accumulator, pid_n, stride_c_n, BLOCK_SIZE_M, BLOCK_SIZE_N)
+    gemm_epilogue(
```
vasqu, line 498: oh ok so we now do the epilogue here as well :D

finegrained-kernels/torch-ext/finegrained_kernels/batched.py
```diff
@@ -1197,11 +1200,15 @@ def w8a8_block_static_fp8_matmul_batched(
...
     )
 
-    As = As.reshape(1).to(torch.float32)
+    # One calibrated scale for the whole matmul, or one per expert -- a MoE calibrates each
+    # separately. Per expert the kernel quantizes in register against the tile's own scale,
+    # because one token routes to several experts whose scales differ and so has no single
+    # pre-quantized form; one scale for all of them pre-quantizes once here instead.
+    As = As.reshape(-1).to(torch.float32)
+    # stride 0 makes every expert read the one calibrated scale; 1 gives each its own
+    as_stride = int(As.numel() == num_experts)
```
vasqu: this loading pattern appears quite often might be a nice util like fn?

vasqu (commented):
Might have found something on quant.py to really recheck but otherwise looks good overall

finegrained-kernels/torch-ext/finegrained_kernels/quant.py
IlyasMoutawwakil: no the decorator should've also been moved, it decorates the wrong thing here, no test caught it tho and that's the real problem, adding a compilation test in test_ops.py catches it.
> kind of weird if this still worked

in eager mode it's a passthrough (to avoid the overhead) so it only fails when it's wrong under torch compile

finegrained-kernels/tests/test_autotuner.py
IlyasMoutawwakil: yes

finegrained-kernels/torch-ext/finegrained_kernels/loading/tiles.py
IlyasMoutawwakil: load_act_static quantizes to e4m3 in-register, so the MMA never sees the raw dtype (i verified bf16, fp16 and fp32 all run). no guard became redundant: the remaining dtype guards are on other paths (weight-only batched upcasts in-loop; dot_scaled needs a bf16/fp16 lhs).

finegrained-kernels/torch-ext/finegrained_kernels/quant.py
IlyasMoutawwakil: to override gatherIdx but moving it to the caller site instead

finegrained-kernels/torch-ext/finegrained_kernels/batched.py
IlyasMoutawwakil: yes

finegrained-kernels/torch-ext/finegrained_kernels/batched.py
IlyasMoutawwakil: good catch, extracted as static_expert_act_operands in quant.py, beside tensor_wide_act_operands

vasqu (approved):
Thanks 🫡

